PR #116, opened 2014-12-19
Popular links

JiggyPete:
Making popular links available through the service call from frontend.
Essentially we expose the 3 most popular articles over the last week in the JSON response.

For this to be deployed we'll need to have a private key (for google analytics api access) setup on the deployment machine. Once deployed, we will want to update the deploy script to invoke `whenever -w` on each deploy to update the cron for the deployment server.


neoeno:
Closing in favour of #120, which includes everything here.

PR #119, opened 2014-12-22
Move linkable object to page model

tomas-stefano:
Refactor LinkableObject.find_page method and move to the right object.

* Here we are reopening the comfortable mexican sofa code to extend
  rather than modifying it.


tomas-stefano:
I'll close this because we're focus on B2b and this method on linkable object needs more improvements than that.

PR #120, opened 2014-12-23
Popular links / Previous & Next links

neoeno:
_Amalgamated from this + #116._

Making popular links available through the service call from frontend.
Essentially we expose the 3 most popular articles over the last week in the JSON response.

For this to be deployed we'll need to have a private key (for google analytics api access) setup on the deployment machine. Once deployed, we will want to update the deploy script to invoke whenever -w on each deploy to update the cron for the deployment server.

---

Exposes the next and previous posts in the article's first category through the article JSON response.

### DIFF
--- a/lib/cms/update_page_views_task.rb
+++ b/lib/cms/update_page_views_task.rb
@@ -6,8 +6,10 @@ class UpdatePageViewsTask
     google_analytics_results = GoogleAnalytics::API.fetch_article_page_views
 
     Comfy::Cms::Page.all_english_articles.each do |article|
-      unless article.update_page_views google_analytics_results
-        log("Error updating article: #{article.id}")
+      begin
+        article.update_page_views(google_analytics_results)
+      rescue
+        log("Error updating article: #{article.id} (#{$ERROR_INFO.message})")
       end
     end
     log('Completed UpdatePageViewsTask')
--- a/lib/cms/update_page_views_task.rb
+++ b/lib/cms/update_page_views_task.rb
@@ -2,10 +2,18 @@ require_relative '../google_analytics/api'
 
 class UpdatePageViewsTask
   def self.run
+    log('Starting UpdatePageViewsTask')
     google_analytics_results = GoogleAnalytics::API.fetch_article_page_views
 
     Comfy::Cms::Page.all_english_articles.each do |article|
-      article.update_page_views google_analytics_results
+      unless article.update_page_views google_analytics_results
+        log("Error updating article: #{article.id}")
+      end
     end
+    log('Completed UpdatePageViewsTask')
+  end
+
+  def self.log(message)
+    Rails.logger.info("#{Time.now}: #{message}")
   end
 end
--- a/lib/cms/update_page_views_task.rb
+++ b/lib/cms/update_page_views_task.rb
@@ -1,0 +1,11 @@
+require_relative '../google_analytics/api'
+
+class UpdatePageViewsTask
+  def self.run
+    google_analytics_results = GoogleAnalytics::API.fetch_article_page_views
+
+    Comfy::Cms::Page.all_english_articles.each do |article|
+      article.update_page_views google_analytics_results
+    end
+  end
+end
--- a/lib/comfortable_mexican_sofa/extensions/page.rb
+++ b/lib/comfortable_mexican_sofa/extensions/page.rb
@@ -1,0 +1,16 @@
+class Comfy::Cms::Page
+  def self.all_english_articles
+    joins(:layout)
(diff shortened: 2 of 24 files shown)


spec/lib/comfortable_mexican_sofa/extensions/page_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+    it 'returns nil when this is the first article' do
+      expect(first_article.previous_page).to be_nil
+    end
+  end
+
+  describe '#next_page' do
+    let(:english_site) { FactoryGirl.create :site }
+    let(:welsh_site) { FactoryGirl.create :site, :welsh }
+    let(:category) { FactoryGirl.create :category }
+    let!(:second_article) { FactoryGirl.create :page, position: 2, site: english_site, categories: [category] }
```
dougdroper: I think I prefer leaving the FactoryGirl namespace, it would be consistent with the frontend and just having 'create' feels like it could be too ambiguous

lib/cms/update_page_views_task.rb
```diff
@@ -0,0 +1,19 @@
+require_relative '../google_analytics/api'
+
+class UpdatePageViewsTask
+  def self.run
+    log('Starting UpdatePageViewsTask')
+    google_analytics_results = GoogleAnalytics::API.fetch_article_page_views
+
+    Comfy::Cms::Page.all_english_articles.each do |article|
+      unless article.update_page_views google_analytics_results
```
dougdroper: I think I would prefer this in a begin rescue block and also raise the exception in the update_page_views method and then log the error here. 
Also can we add parentheses around the method arguments for consistency.

lib/comfortable_mexican_sofa/extensions/page.rb
```diff
@@ -0,0 +1,53 @@
...
+    joins(:layout)
+    .joins(:site)
+    .where(
+      comfy_cms_sites: { label: 'en' },
+      comfy_cms_layouts: { identifier: 'article' }
+    )
+  end
+
+  def update_page_views(analytics)
+    matching_analytic = analytics.find { |analytic| analytic[:label] == slug }
```
dougdroper: Can we move this line to a method?

spec/lib/comfortable_mexican_sofa/extensions/page_spec.rb
tomas-stefano: Ok. But there are some other places (models, presenters and controllers) that are using just create. So if we accept we need to change it, after, for consistency.

lib/comfortable_mexican_sofa/extensions/page.rb
```diff
@@ -0,0 +1,53 @@
...
+
+    category = categories.first
+    site.pages.for_category(category.label)
+  end
+
+  def fetch_page(pages, page_index)
+    return nil if page_index < 0 || page_index >= pages.length
+    pages[page_index]
+  end
+end
```
neoeno: Good call. Actioned.

spec/lib/comfortable_mexican_sofa/extensions/page_spec.rb
neoeno: Seems like we're saying 'leave as-is' for now? Is that the consensus? Not sure if I'm reading correctly.

benlovell: I vote `create`. We all know exactly where it comes from and it's the general standard everywhere else.

spec/serializers/page_serializer_spec.rb
```diff
@@ -0,0 +1,97 @@
...
+          .with(3)
+          .and_return([english_article])
+
+        popular_links = subject.related_content[:popular_links]
+        expect(popular_links).to eq([
+          { title: 'First Article in Welsh', path: '/cy/articles/first-article-in-welsh' }
+        ])
+      end
+
+    end
```
neoeno: Thanks for this! You'll notice we've included most of it wholesale ;)

One fun thing we came up against was that this:

```ruby
subject(:popular_links) do
  subject.related_content[:popular_links]
end
```

Caused a segfault. Didn't investigate, but guessing due to smashing the stack.

We changed it to a let and it worked fine however. Same with the other chunk. Thanks again!

tomas-stefano: Yeah, fun indeed :+1:

neoeno:
Ready for merge I think.

tomas-stefano:
lgtm :+1: But we need one more thumbs up.

PR #122, opened 2014-12-30
B2b creating corporate page

tomas-stefano:
Three things:

* Refactor one method of Hippo Importer (is not related to the feature development, but was related to investigate the actual structure of Hippo articles)
* Remove layouts from menu like the P.O requested.
* Adding a rake task to create the corporate article type.

### DIFF
--- a/lib/cms/hippo_import.rb
+++ b/lib/cms/hippo_import.rb
@@ -68,13 +68,9 @@ class HippoImport
   def import!
     records.select { |r| next_docs.include?(r.send(article_id)) }.map do |record|
       puts "Importing: #{record.id}"
-      page = if @_site == 'en'
-               Comfy::Cms::Page.where(slug: record.id, site: site).first
-             else
-               Comfy::Cms::Page.where(translation_id: record.translation_id, site: site).first
-             end
+      page = find_page(record) || Comfy::Cms::Page.new
 
-      (page || Comfy::Cms::Page.new).tap do |p|
+      page.tap do |p|
         p.site = site
         p.layout = layout
         p.parent = parent
@@ -91,6 +87,16 @@ class HippoImport
         ]
         p.save unless page && page.state == 'published'
       end
+    end
+  end
+
+  private
+
+  def find_page(record)
+    if @_site == 'en'
+      Comfy::Cms::Page.where(slug: record.id, site: site).first
+    else
+      Comfy::Cms::Page.where(translation_id: record.translation_id, site: site).first
     end
   end
 end
--- a/lib/tasks/b2b.rake
+++ b/lib/tasks/b2b.rake
@@ -1,0 +1,19 @@
+namespace :b2b do
+  desc 'Create corporate article type'
+  task create_type: :environment do
+    ActiveRecord::Base.logger = Logger.new(STDOUT)
+
+    site = Comfy::Cms::Site.first
+    layout = site.layouts.find_or_initialize_by(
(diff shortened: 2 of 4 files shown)


dougdroper:
:+1:

lib/cms/hippo_import.rb
```diff
@@ -93,4 +89,12 @@ def import!
       end
     end
   end
+
+  def find_page(record)
```
asmega: should this be a private method? or does it need to be publicly exposed?

lib/tasks/b2b.rake
```diff
@@ -0,0 +1,19 @@
...
+      label: 'Corporate Article',
+      identifier: 'corporate_article',
+      content: "{{ cms:page:content:rich_text }}\n"
+    )
+
+    Rails.logger.info("*" * 80)
+    Rails.logger.info("Saving ##{layout.identifier} for site ##{site.label} and mirroring to others sites.")
+    Rails.logger.info("*" * 80)
+
+    Rails.logger.info("Layout saved successfuly. Details: \n#{layout.inspect}.\nMirrors: #{layout.mirrors}") if layout.save
```
asmega: typo on successfuly should be successfully

lib/cms/hippo_import.rb
tomas-stefano: Good one. Should be private. I will add to it.

lib/tasks/b2b.rake
```diff
@@ -0,0 +1,19 @@
+namespace :b2b do
+  desc 'Create corporate article type'
+  task create_type: :environment do
+    Rails.logger = ActiveRecord::Base.logger = Logger.new(STDOUT)
```
benlovell: If you're redirecting the rails logger to `STDOUT` you may as well just `puts` these messages.

lib/tasks/b2b.rake
benlovell: It irks me slightly when I see a long line and then a surprise statement modifier at the end. I've heard them called 'if surprises' before. Which seems pretty apt to me. 

One way to solve this is by instead saying `layout.save && Rails.logger.info(...)` but you might find that offensive too depending on your leanings.

lib/tasks/b2b.rake
tomas-stefano: Yeah, I always use this type of thing and always put some file to be write the log. Then I changed to STDOUT because was not so critical. I will use puts.

lib/tasks/b2b.rake
tomas-stefano: Actually, this is a great idea. I will do it.

tomas-stefano:
thanks @asmega and @benlovell. Now I think it's okay.

lib/tasks/b2b.rake
```diff
@@ -0,0 +1,19 @@
+namespace :b2b do
+  desc 'Create corporate article type'
+  task create_type: :environment do
+    ActiveRecord::Base.logger = Logger.new(STDOUT)
```
benlovell: Whoops. This can go now too.

tomas-stefano: Actually I prefer to see the select/insert SQL statements in the output when I am running the task (even with the puts more humanized).

benlovell: Cool. Good point.

benlovell:
LGTM! :+1:

PR #123, opened 2015-01-08
adding latest blog post links (form Publify) to page_serializer for frontend

JiggyPete:
Note: 
  'PUBLIFY_HOSTNAME' is set in .env for development.
  For QA & production we set this up in the Scripts repo.

### DIFF
--- a/.env.dev
+++ b/.env.dev
@@ -1,3 +1,5 @@
 FRONTEND_URL="localhost:5000"
 GA_API_EMAIL_ADDRESS="google.service.account@example.com"
 GA_PRIVATE_KEY_PATH="path/to/private.key"
+PUBLIFY_HOSTNAME="blog.moneyadviceservice.org.uk"
+PUBLIFY_PORT="80"
--- a/lib/publify/api.rb
+++ b/lib/publify/api.rb
@@ -1,0 +1,14 @@
+require 'net/http'
+
+module Publify
+  class API
+    def self.latest_links(limit)
+      uri = URI("http://#{PUBLIFY_HOSTNAME}/articles.json")
+      response = Net::HTTP.get uri
+      JSON.parse(response).first(limit)
+    rescue => e
+      Rails.logger.error(e.message)
+      []
+    end
+  end
+end
--- a/lib/publify/api.rb
+++ b/lib/publify/api.rb
@@ -3,7 +3,7 @@ require 'net/http'
 module Publify
   class API
     def self.latest_links(limit)
-      uri = URI("http://#{PUBLIFY_HOSTNAME}/articles.json")
+      uri = URI("http://#{ENV['PUBLIFY_HOSTNAME']}/articles.json")
       response = Net::HTTP.get uri
       JSON.parse(response).first(limit)
     rescue => e
--- a/lib/publify/api.rb
+++ b/lib/publify/api.rb
@@ -3,9 +3,11 @@ require 'net/http'
 module Publify
   class API
     def self.latest_links(limit)
-      uri = URI("http://#{ENV['PUBLIFY_HOSTNAME']}/articles.json")
-      response = Net::HTTP.get uri
-      JSON.parse(response).first(limit)
+      connection = Net::HTTP.new(ENV['PUBLIFY_HOSTNAME'], ENV['PUBLIFY_PORT'].to_i)
+      connection.open_timeout = 2
+      connection.read_timeout = 2
+      response = connection.get('/articles.json')
+      JSON.parse(response.body).first(limit)
     rescue => e
       Rails.logger.error(e.message)
       []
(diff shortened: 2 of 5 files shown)


lib/publify/api.rb
```diff
@@ -0,0 +1,18 @@
...
+        response = Net::HTTP.get uri
+        JSON.parse(response).first(limit)
+      rescue Exception => e
+        Rails.logger.error(e.message)
+        []
+      end
+    end
+
+  end
+end
```
tomas-stefano: Humm, I know there are just one resource for now, but I prefer organize this in a resource way (e.g):

```ruby
   module Publify
     class Article
       include ActiveModel::Model
       attr_accessor :title, :link
       def self.resource_name
         'articles.json'
       end

       def self.latest(limit)
         # run_request(resource_name).map { |article_hash| new(article_hash) }
       end
     end
   end
```

You get the idea. Wdyt?

JiggyPete: Looks nice yet seems like wrapping it up in an object that, at the minute, we don't need.

lib/publify/api.rb
```diff
@@ -0,0 +1,14 @@
+require 'net/http'
+
+module Publify
+  class API
+    def self.latest_links(limit)
+      uri = URI("http://#{PUBLIFY_HOSTNAME}/articles.json")
+      response = Net::HTTP.get uri
```
neoeno: Looking at the [docs](http://ruby-doc.org/stdlib-2.1.1/libdoc/net/http/rdoc/Net/HTTP.html#class-Net::HTTP-label-GET), seems this could be written as:

```ruby
response = Net::HTTP.get(PUBLIFY_HOSTNAME, '/articles.json')
```

neoeno: Also, do we know what the timeout is here? It should probably be short to avoid a stalled publify (or heroku issue) causing cms/frontend to be unacceptably sluggish.

spec/serializers/page_serializer_spec.rb
```diff
@@ -6,10 +6,44 @@
...
+      it 'maps the 3 latest links' do
+        result = subject.related_content[:latest_blog_post_links]
+        expect(result.length).to eq(3)
+      end
+
+      it 'maps the latest links correctly' do
+        result = subject.related_content[:latest_blog_post_links]
+
+        expect(result.first).to eq(
+          {
```
neoeno: Rubocop says:

```
spec/serializers/page_serializer_spec.rb:37:11: C: Redundant curly braces around a hash parameter.
          {
          ^
```

lib/publify/api.rb
benlovell: Yep, could we make sure this isn't unbounded. Also, how is this being called? Is it frequent, is this response cached?

JiggyPete: @neoeno We're not referring to the PUBLIFY_HOSTNAME correctly which is causing the issue.

@benlovell Can you clarify what you mean by "unbounded"? Currently the call is made for each article that is rendered on frontend.

neoeno: Ay, that's flagged above. This comment was about timeouts.

So, after looking into it, I discovered that by default `Net::HTTP.get` will wait forever to get a connection, and then 60 seconds for a response. Seems like we should set a shorter timeout to avoid issues with Publify cascading into a total loss of service.

lib/publify/api.rb
```diff
@@ -0,0 +1,14 @@
+require 'net/http'
+
+module Publify
+  class API
+    def self.latest_links(limit)
+      uri = URI("http://#{PUBLIFY_HOSTNAME}/articles.json")
```
neoeno: We should probably add it in `env.dev` too.

.env.dev
```diff
@@ -1,3 +1,4 @@
 FRONTEND_URL="localhost:5000"
 GA_API_EMAIL_ADDRESS="google.service.account@example.com"
 GA_PRIVATE_KEY_PATH="path/to/private.key"
+PUBLIFY_HOSTNAME="localhost:4000"
```
neoeno: Might be nice if publify defaulted to this locally (or, at least, something that's not :3000).

@baphled wyt?

JiggyPete: looks good

lib/publify/api.rb
```diff
@@ -0,0 +1,16 @@
+require 'net/http'
+
+module Publify
+  class API
+    def self.latest_links(limit)
+      connection = Net::HTTP.new(ENV['PUBLIFY_HOSTNAME'], ENV['PUBLIFY_PORT'].to_i)
+      connection.open_timeout = 2
+      connection.read_timeout = 2
+      response = connection.get('/articles.json')
+      JSON.parse(response.body).first(limit)
```
neoeno: FAO @benlovell.

benlovell:
:+1:

tomas-stefano:
:+1:

PR #124, opened 2015-01-13
Adds drop shadow underneath fixed toolbar

ourmaninamsterdam:
Makes it easier to see the toolbar when it is fixed.


dougdroper:
:+1: :+1:

PR #125, opened 2015-01-14
Fix image aspect ratios on content grid

ourmaninamsterdam:
Previously the images were being stretched to fit a 1:1 aspect ratio. This fix makes them fit to their native aspect ratios within the content grid.


dougdroper:
:+1:

PR #126, opened 2015-01-14
Fix country flag image layout

ourmaninamsterdam:
The country toggle was wrapping beneath the country flag image. This floats it to the right of the toggle.


dougdroper:
:+1:

PR #129, opened 2015-01-14
Hippo import

tomas-stefano:


lib/cms/hippo_importer/base.rb
```diff
@@ -0,0 +1,99 @@
...
+      def find_page(record)
+        if @_site == 'en'
+          Comfy::Cms::Page.where(slug: record.id, site: site).first
+        else
+          Comfy::Cms::Page.where(translation_id: record.translation_id, site: site).first
+        end
+      end
+    end
+  end
+end
```
tomas-stefano: I moved the old class to this but with a different name and hash options in the initialize. I would like to refactor this entirely, but is not inside of our scope. @dougdroper if we refactor that in a near future, what do you think about it separate this classes in: one class is the real importer and the other is the parser?

lib/tasks/importer.rake
```diff
@@ -1,20 +1,19 @@
...
+      type  = ENV['type'] || 'base'
 
       puts "Importing CMS Data from file [#{from}]..."
 
       file = File.read(File.expand_path(from))
       docs = File.read(File.expand_path(docs)).split("\n") if docs
 
-      imported = HippoImport.new(file, docs, to).import!
+      import = "Cms::HippoImporter::#{type.classify}".constantize.new(data: file, docs: docs, to: to)
+      imported = import.import!
```
tomas-stefano: with this you can run the corporate pages with:

`rake cms:hippo:import TYPE=corporate_page ...`

lib/cms/hippo_importer/base.rb
tomas-stefano: But the only improvement that I made is that you can overwrite the hippo type and the layout on the subclass. So \o/

dougdroper: Yes this file has been a bit of a dumping ground for things, in the hope that it would be thrown away after the initial import. But seems like it is sticking. Definitely happy with splitting the parsing from the importing.

dougdroper:
:+1:

tomas-stefano:
One more thumbs-up cc @benlovell @finitud :)

lib/cms/hippo_importer/base.rb
```diff
@@ -0,0 +1,99 @@
+module Cms
+  module HippoImporter
+    class Base
+      def initialize(data:, docs:, to: 'en', parser: HippoXmlParser, html_decoder: HTMLEntities.new)
+        @records = parser.parse(data, [hippo_type])
+        @_site = to || 'en'
```
benlovell: The `|| 'en'` is not needed here.

lib/cms/hippo_importer/base.rb
```diff
@@ -0,0 +1,99 @@
...
+        @docs = docs || []
```
benlovell: I'd move the `[]` to default arg for `docs` instead.

tomas-stefano: The old code was wrong and I didn't saw. Good catch.

lib/cms/hippo_importer/base.rb
```diff
@@ -0,0 +1,99 @@
...
+      def site
+        @site ||= Comfy::Cms::Site.find_by(label: @_site)
+      end
+
+      def hippo_type
+        'contentauthoringwebsite:Guide'
+      end
+
+      def layout
+        @_layout ||= (site.try(:layouts) || []).first
```
benlovell: You could tidy this up as `Array(site.try(:layouts)).first`

lib/cms/hippo_importer/base.rb
benlovell: What's with the underscore prefix on these variables? Is it to avoid naming clashes?

lib/cms/hippo_importer/base.rb
tomas-stefano: Acctually, the old code is wrong. All the site has layouts. This hippo importer should not have this. I will use just site.layouts.find_by(identifier: 'article') # because  this should be default. Thanks

lib/cms/hippo_importer/base.rb
```diff
@@ -0,0 +1,99 @@
...
+          e.replace('({' + e.attributes['src'].value.gsub('https://www.youtube.com/embed/', '') + '})')
+        end
+      end
+
+      def docs
+        @cached_docs ||= @docs.empty? ? Comfy::Cms::Page.all.map(&:slug) : @docs
+      end
+
+      def next_docs
+        @next_docs ||= @_site == 'en' ? docs : docs.map { |a| Comfy::Cms::Page.find_by(slug: a) }.map(&:translation_id)
```
benlovell: The `@_site == 'en'` comparison is done in enough places I'd move it into a predicate, perhaps `english?`

lib/cms/hippo_importer/corporate_page.rb
```diff
@@ -0,0 +1,13 @@
+module Cms
+  module HippoImporter
+    class CorporatePage < HippoImporter::Base
+      def hippo_type
+        'contentauthoringwebsite:Guide'
+      end
+
+      def layout
+        @_layout ||= site.layouts.find_by(identifier: 'corporate_article')
```
benlovell: Same here with the underscored variable.

lib/tasks/importer.rake
```diff
@@ -1,20 +1,19 @@
...
   namespace :hippo do
-    desc 'Import Hippo XML data into database (options: FROM=folder_name TO=site_identifier DOCS=file_of_docs)'
-
-    task :import => :environment do
+    desc 'Import Hippo XML data into database \n(options: FROM=folder_name TO=site_identifier DOCS=file_of_docs TYPE=type)'
+    task import: :environment do
       from  = ENV['FROM']
       to    = ENV['TO']
       docs  = ENV['DOCS']
+      type  = ENV['type'] || 'base'
```
benlovell: I tend to prefer `ENV.fetch('type', 'base')` but I know that's personal preference.

lib/cms/hippo_importer/base.rb
tomas-stefano: I had the same questions. I just move the old code to a other location that I thought that was better. I will change that too.

tomas-stefano:
I think now I made all the changes. cc @dougdroper and @benlovell 
I move some code that is not responsability of the importer to a class called HippoContent that makes the conversion to 'contento'.

lib/cms/hippo_importer/base.rb
```diff
@@ -0,0 +1,86 @@
...
+    class Base
+      attr_reader :records
+
+      def initialize(data:, docs: [], to: 'en', parser: HippoXmlParser)
+        @records = parser.parse(data, [hippo_type])
+        @_site = to
+        @docs = docs
+      end
+
+      def hippo_type
```
benlovell: I'd make this a constant instead. Also accessing this property from the initializer is a little strange. That's what made me do a double take.

lib/cms/hippo_importer/base.rb
```diff
@@ -0,0 +1,86 @@
...
+        'contentauthoringwebsite:Guide'
+      end
+
+      def site
+        @site ||= Comfy::Cms::Site.find_by(label: @_site)
```
benlovell: I'd move this to `initialize` and then you don't need to mess around with the whole `@_site` and `@site` stuff.

lib/cms/hippo_importer/base.rb
benlovell: Ah, actually I see this is a 'template method'

tomas-stefano: Yes, agreed in parts. Following this idea, I will extract the line from initialize to a method called records and remove the attr_reader records. But the hippo_type is needed, because the subclass will overwrite, because there are different types on hippo corresponding to a different layouts on our database. Like the CorporatePage class for example.

benlovell: See my comment above :wink:

tomas-stefano: Cool. It was on the same time that I was commenting, ;)

tomas-stefano:
I think now is everything ok.

tomas-stefano:
My feelings on understand the hippo structure, the hippo xml content, the hippo xml parser, the hippo import and change all the code. :) :)

feeling-legacy-code

benlovell:
:+1:

PR #131, opened 2015-01-15
Exposing related_links for frontend

JiggyPete:


app/models/comfy/cms/page.rb
```diff
@@ -3,14 +3,28 @@
 class Comfy::Cms::Page < ActiveRecord::Base
   scope :most_popular, -> (number_of_items) { unscoped.order('page_views desc').limit(number_of_items) }
 
-  def self.all_english_articles
+  scope :with_tags, -> (tags) { joins(:taggings).where('tag_id IN (?)', tags.map(&:id)) }
+
+  scope :remove_self_from_results, -> (page) { where.not(id: page.id) }
+
+  scope :remove_alternate_language_page_when_welsh, (lambda do |page|
```
tomas-stefano: Here are using the lambda keywords and on top are using ->.

app/models/comfy/cms/page.rb
```diff
@@ -3,14 +3,28 @@
...
+    where.not(id: page.mirrors.first.try(:id))
```
tomas-stefano: If page.mirrors.first.try(:id) is nil they will make an expected SQL query, right? What happens if this is nil?

app/models/comfy/cms/page.rb
```diff
@@ -3,14 +3,28 @@
...
+  end)
+
+  scope :sort_by_tag_similarity, (lambda do |keywords|
```
tomas-stefano: Same lambda and -> said more on top.

app/models/comfy/cms/page.rb
```diff
@@ -3,14 +3,28 @@
...
+    with_tags(keywords)
+      .group('comfy_cms_pages.id')
+      .order('COUNT(*) DESC, comfy_cms_pages.page_views DESC')
+  end)
+
+  scope :all_english_articles, (lambda do
```
tomas-stefano: Same lambda and -> said more on top.

app/serializers/page_serializer.rb
```diff
@@ -32,6 +33,13 @@ def latest_blog_post_links
     end
   end
 
+  def related_links
+    object.related_links(3).map do |article|
+      article = article.mirrors.first if current_locale == 'cy'
+      build_json_link article
+    end
+  end
+
```
tomas-stefano: The private build_json_link method, refers to page more than self. I think we can remove this method and replaces and move the code to another class. Maybe could be PageLinkSerializer with title and path as attribute.

app/serializers/page_serializer.rb
```diff
@@ -32,6 +33,13 @@ def latest_blog_post_links
     end
   end
 
+  def related_links
+    object.related_links(3).map do |article|
+      article = article.mirrors.first if current_locale == 'cy'
+      build_json_link article
+    end
+  end
```
tomas-stefano: This related links is the same behaviour as popular links. Should extracted to a method.

app/models/comfy/cms/page.rb
neoeno: Not working on this today I think, but I was curious so I ran something to find out:

```
> Comfy::Cms::Page.all.unscoped.where.not(id: nil)
  Comfy::Cms::Page Load (11.1ms)  SELECT `comfy_cms_pages`.* FROM `comfy_cms_pages`  WHERE (`comfy_cms_pages`.`id` IS NOT NULL)
```

Shouldn't change anything, but it is admittedly unexpected.

app/models/comfy/cms/page.rb
neoeno: You probably already went over this in person, but this was the effect of the 'max line length' rubocop rule, and the 'Use the `lambda` method for multi-line lambdas.' rule. So if a lambda was over a certain length, it needed to be broken onto a couple of lines, and then needed to use the keyword syntax.

app/models/comfy/cms/page.rb
```diff
@@ -3,14 +3,28 @@
 class Comfy::Cms::Page < ActiveRecord::Base
   scope :most_popular, -> (number_of_items) { unscoped.order('page_views desc').limit(number_of_items) }
 
-  def self.all_english_articles
+  scope :with_tags, -> (tags) { joins(:taggings).where('tag_id IN (?)', tags.map(&:id)) }
```
neoeno: Suggestion:

```scope :with_tags, -> (tags) { joins(:taggings).where('taggings.tag_id' => tags.map(&:id)) }```

This removes a little SQL from the ruby, and clarifies both to the reader and in the SQL that `tag_id` is on `taggings` — currently if someone went wild and added `tag_id` to the pages table this would break. No one will ever do that, but still.

Considered switching `tags.map(&:id)` to `tags.pluck('id')`, which would be smarter, but only works with ActiveRecord relations, so I figured it was a bad idea since you might want to pass in an array.

File: app/models/comfy/cms/page.rb (hunk 2 above)
Comment (neoeno) [inline app/models/comfy/cms/page.rb]: I've also been worrying since I wrote this that it might be important to exclude all the mirrors? Not just the first one arbitrarily?

File: app/models/comfy/cms/page.rb [hunk 8]
```diff
@@ -3,14 +3,28 @@
...
+  scope :remove_self_from_results, -> (page) { where.not(id: page.id) }
+
+  scope :remove_alternate_language_page_when_welsh, (lambda do |page|
+    where.not(id: page.mirrors.first.try(:id))
+  end)
+
+  scope :sort_by_tag_similarity, (lambda do |keywords|
+    with_tags(keywords)
+      .group('comfy_cms_pages.id')
+      .order('COUNT(*) DESC, comfy_cms_pages.page_views DESC')
```
neoeno: It might be good to comment this to explain what's going on. If I see `COUNT(*)` I think 'number of records', but its meaning is defined by the GROUP BY here so it'd be nice to signal that.

app/models/comfy/cms/page.rb
```diff
@@ -3,14 +3,28 @@
...
+
+  scope :all_english_articles, (lambda do
     joins(:layout)
     .joins(:site)
     .where(
       comfy_cms_sites: { label: 'en' },
       comfy_cms_layouts: { identifier: 'article' }
     )
-  end
+  end)
```
neoeno: It'd be nice to see the indentation in this file consistent, e.g.:

```ruby
  scope :all_english_articles, (lambda do
    joins(:layout)
      .joins(:site)
      .where(
        comfy_cms_sites: { label: 'en' },
        comfy_cms_layouts: { identifier: 'article' }
      )
  end)
```

app/models/comfy/cms/page.rb
benlovell: on L6 I'm pretty sure you could use `joins(:taggings).where(tag_id: tag_ids)` 

It has the smarts to use `in` if it's an array.

neoeno: !!! I just remembered we tried to do it this way on Wednesday and it didn't work. It's a really tough one to debug because `Page#mirrors` isn't available in `rails c` for some reason?

But the root of it is that when you run a query on your own ActiveRecord class in a scope, it executes the query _within all your chained scopes_. So when `#mirrors` queries `Page`, it'll tag all the existing criteria on and you won't get the right mirrors.

If you want to see for yourself, look at the SQL that is generated when you run `#related_links`:

```
Q1: Tag Load (1.1ms)  SELECT `tags`.* FROM `tags` INNER JOIN `taggings` ON `tags`.`id` = `taggings`.`tag_id` WHERE `taggings`.`taggable_id` = 605 AND `taggings`.`taggable_type` = 'Comfy::Cms::Page'

Q2: Comfy::Cms::Site Load (0.2ms)  SELECT `comfy_cms_sites`.* FROM `comfy_cms_sites`  WHERE `comfy_cms_sites`.`is_mirrored` = 1

Q3: Comfy::Cms::Page Load (0.4ms)  SELECT  `comfy_cms_pages`.* FROM `comfy_cms_pages` INNER JOIN `comfy_cms_layouts` ON `comfy_cms_layouts`.`id` = `comfy_cms_pages`.`layout_id` INNER JOIN `comfy_cms_sites` ON `comfy_cms_sites`.`id` = `comfy_cms_pages`.`site_id` INNER JOIN `taggings` ON `taggings`.`taggable_id` = `comfy_cms_pages`.`id` AND `taggings`.`taggable_type` = 'Comfy::Cms::Page' WHERE `comfy_cms_sites`.`label` = 'en' AND `comfy_cms_layouts`.`identifier` = 'article' AND `taggings`.`tag_id` IN (2, 10, 70) AND (`comfy_cms_pages`.`id` != 605) AND `comfy_cms_pages`.`site_id` = 2 AND `comfy_cms_pages`.`full_path` = '/do-you-need-to-borrow-money' GROUP BY comfy_cms_pages.id  ORDER BY COUNT(*) DESC, comfy_cms_pages.page_views DESC LIMIT 1

Q4: Comfy::Cms::Page Load (4.8ms)  SELECT  `comfy_cms_pages`.* FROM `comfy_cms_pages` INNER JOIN `comfy_cms_layouts` ON `comfy_cms_layouts`.`id` = `comfy_cms_pages`.`layout_id` INNER JOIN `comfy_cms_sites` ON `comfy_cms_sites`.`id` = `comfy_cms_pages`.`site_id` INNER JOIN `taggings` ON `taggings`.`taggable_id` = `comfy_cms_pages`.`id` AND `taggings`.`taggable_type` = 'Comfy::Cms::Page' WHERE `comfy_cms_sites`.`label` = 'en' AND `comfy_cms_layouts`.`identifier` = 'article' AND `taggings`.`tag_id` IN (2, 10, 70) AND (`comfy_cms_pages`.`id` != 605) AND (`comfy_cms_pages`.`id` IS NOT NULL) GROUP BY comfy_cms_pages.id  ORDER BY COUNT(*) DESC, comfy_cms_pages.page_views DESC LIMIT 3
```

Q3 fetches the mirrors, Q4 fetches the related links. If you scroll along you'll see they're virtually identical, except that right at the end of Q3 is the criteria for finding mirrors.

spec/models/comfy/cms/page_spec.rb
```diff
@@ -1,5 +1,78 @@
...
+        expect(animals_article.related_links(2)).to eq([related_popular_article, related_unpopular_article])
+      end
+    end
+
+    context 'when finding a welsh article' do
+      it 'we should exclude the alternate english version from the results' do
+        welsh_animal_article = create :welsh_article, keywords: [animals_tag], page_views: 2
+        related_article = create :english_article, keywords: [animals_tag], page_views: 1
+
+        allow(welsh_animal_article).to receive(:mirrors).and_return([animals_article])
```
neoeno: Related to the mirrors stuff, but this test passes if you comment out this line.

app/models/comfy/cms/page.rb
neoeno: Actioned.

neoeno:
@tomas-stefano @benlovell wyt?

spec/serializers/page_serializer_spec.rb
```diff
@@ -42,9 +43,42 @@
...
+          site.label = 'en'
+        end
+
+        it 'returns english related links' do
+          expect(related_links).to eq([
+            { title: 'First Article', path: '/en/articles/first-article' },
+            { title: 'Second Article', path: '/en/articles/second-article' }
+          ])
+        end
+      end
```
tomas-stefano: What happen when most related articles is not english? Is there a spec needed here?

spec/models/comfy/cms/page_spec.rb
```diff
@@ -1,45 +1,87 @@
...
     end
-  end
 
-  describe '#next_page' do
-    let(:english_site) { create :site }
-    let(:welsh_site) { create :site, :welsh }
-    let(:category) { create :category }
-    let!(:second_article) { create :page, position: 2, site: english_site, categories: [category] }
+    context 'when finding a welsh article' do
+      it 'we should exclude the alternate english version from the results' do
```
tomas-stefano: You could remove 'we should' from here

spec/models/comfy/cms/page_spec.rb
```diff
@@ -1,45 +1,87 @@
...
+        layout = create :layout
```
tomas-stefano: This could be put on a let.

spec/models/comfy/cms/page_spec.rb
```diff
@@ -1,45 +1,87 @@
...
+               keywords:   [animals_tag],
+               page_views: 5,
+               full_path:  '/animals',
+               layout:     layout
+
+        welsh_animal_article = create :welsh_article,
+                                      keywords:   [animals_tag],
+                                      page_views: 2,
+                                      full_path:  '/animals',
+                                      layout:     layout
```
tomas-stefano: This two could be on a let.

spec/serializers/page_serializer_spec.rb
neoeno: Agreed — can see it being done in other tests so it's now done here.

benlovell:
Seems legit :+1:

tomas-stefano:
:+1:

PR #136, opened 2015-01-22
Suppress related and popular links

JiggyPete:
This PR involves providing a checkbox in the Contento UI to suppress and article from being returned in the Related and Popular Links json.

When marking an article as suppressed, it's mirror in other languages should also be suppressed.

### DIFF
--- a/app/models/comfy/cms/page.rb
+++ b/app/models/comfy/cms/page.rb
@@ -1,7 +1,14 @@
 require ComfortableMexicanSofa::Engine.root.join('app', 'models', 'comfy', 'cms', 'page.rb')
 
 class Comfy::Cms::Page < ActiveRecord::Base
-  scope :most_popular, -> (number_of_items) { unscoped.order('page_views desc').limit(number_of_items) }
+  scope :ignore_suppressed, -> { where(suppress_from_links_recirculation: false) }
+
+  scope :most_popular, (lambda do |number_of_items|
+    unscoped
+      .order('page_views desc')
+      .ignore_suppressed
+      .limit(number_of_items)
+  end)
 
   scope :with_tags, -> (tags) { joins(:taggings).where('taggings.tag_id' => tags.map(&:id)) }
 
@@ -37,9 +44,16 @@ class Comfy::Cms::Page < ActiveRecord::Base
     Comfy::Cms::Page
       .unscoped
       .all_english_articles
+      .ignore_suppressed
       .sort_by_tag_similarity(keywords)
       .remove_self_from_results(self)
       .where.not(id: mirrored_page.try(:id))
       .limit(limit)
+  end
+
+  def suppress_mirrors_from_links_recirculation
+    mirrors.each do |mirror|
+      mirror.update_column(:suppress_from_links_recirculation, suppress_from_links_recirculation)
+    end
   end
 end
--- a/spec/controllers/pages_controller_spec.rb
+++ b/spec/controllers/pages_controller_spec.rb
@@ -137,120 +137,135 @@ RSpec.describe PagesController do
       }
     end
 
-    before do
-      put :update,
-          site_id:     site.id,
-          id:          page.id,
-          state_event: state_event,
-          page:        page_attributes
-    end
-
-    context 'when passes the "save_unsaved" event state from an "unsaved" page' do
+    context 'suppress_mirrors_from_links_recirculation on requested page ' do
       let!(:page) { create(:page, state: 'unsaved') }
-      let(:state_event) { 'save_unsaved' }
 
       it 'persists page as "draft"' do
-        expect(assigns[:page].state).to eq 'draft'
+        expect_any_instance_of(Comfy::Cms::Page).to receive(:suppress_mirrors_from_links_recirculation)
+
+        put :update,
+            site_id:     site.id,
+            id:          page.id,
+            state_event: 'save_unsaved',
+            page:        page_attributes
       end
     end
 
-    context 'when passes the same event state from the page' do
-      let!(:page) { create(:page, state: 'published') }
-      let(:state_event) { 'publish' }
-
-      it 'does not save any revisions' do
-        expect(assigns[:page].revisions).to eq []
+    context 'when a request is made' do
+      before do
+        put :update,
+            site_id:     site.id,
+            id:          page.id,
+            state_event: state_event,
+            page:        page_attributes
       end
-    end
+      context 'when passes the "save_unsaved" event state from an "unsaved" page' do
+        let!(:page) { create(:page, state: 'unsaved') }
+        let(:state_event) { 'save_unsaved' }
 
-    context 'when passes the "publish" event state' do
-      let!(:page) { create(:page, state: 'draft') }
-      let(:state_event) { 'publish' }
-
-      it 'persists page as "published"' do
-        expect(assigns[:page].state).to eq 'published'
-      end
-
-      context 'creating revision' do
-        before do
-          assigns[:page].revisions.reload
-          expect(last_revision).to_not be nil
-        end
-
-        it 'saves the revision event and author' do
-          expect(last_revision.data.symbolize_keys).to eq(
-            event: 'published',
-            previous_event: 'draft',
-            blocks_attributes: [],
-            author: {
-              id: current_user.id,
-              name: current_user.name
-            }
-          )
+        it 'persists page as "draft"' do
+          expect(assigns[:page].state).to eq 'draft'
         end
       end
-    end
 
-    context 'when changing the block attributes' do
-      let!(:page) { create(:page, state: 'published') }
-      let(:state_event) { 'publish' }
-      let(:page_attributes) do
-        {
-          label: 'Another label',
-          slug:  'another-slug',
-          blocks_attributes: {
-            '0' => {
-              content:    'block-content',
-              identifier: 'content'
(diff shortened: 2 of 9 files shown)


spec/models/comfy/cms/page_spec.rb
```diff
@@ -1,5 +1,57 @@
 RSpec.describe Comfy::Cms::Page do
 
+  describe 'suppress_from_links_recirculation' do
```
tomas-stefano: I think you missed the '#'

app/models/comfy/cms/page.rb
```diff
@@ -37,9 +44,16 @@ def related_links(limit)
...
       .all_english_articles
+      .ignore_suppressed
       .sort_by_tag_similarity(keywords)
       .remove_self_from_results(self)
       .where.not(id: mirrored_page.try(:id))
       .limit(limit)
   end
+
+  def suppress_mirrors_from_links_recirculation
+    mirrors.each do |m|
```
tomas-stefano: Can you renamed the variable to mirror?

app/models/comfy/cms/page.rb
```diff
@@ -37,9 +44,16 @@ def related_links(limit)
...
+      m.update_column(:suppress_from_links_recirculation, suppress_from_links_recirculation)
+    end
+  end
```
tomas-stefano: Btw, is the comfy not mirroring automatically the suppress_from_links_recirculation? Do we need this method?

spec/models/comfy/cms/page_spec.rb
```diff
@@ -1,5 +1,57 @@
...
+             site: welsh_site
+    end
+
+    context 'update welsh article when updating english_article' do
+      it 'to be suppressed' do
+        english_article.suppress_from_links_recirculation = true
+        english_article.suppress_mirrors_from_links_recirculation
+        welsh_article.reload
+
+        expect(welsh_article.suppress_from_links_recirculation).to be_truthy
```
tomas-stefano: The suppress_from_links_recirculation is a boolean. You can do  expect(welsh_article).to be_suppress_from_links_recirculation

spec/models/comfy/cms/page_spec.rb
```diff
@@ -1,5 +1,57 @@
...
+
+        expect(english_article.suppress_from_links_recirculation).to be_truthy
+      end
+
+      it 'to not be suppressed' do
+        welsh_article.suppress_from_links_recirculation = false
+        welsh_article.suppress_mirrors_from_links_recirculation
+        english_article.reload
+
+        expect(english_article.suppress_from_links_recirculation).to be_falsey
```
tomas-stefano: Same about comment above. `The suppress_from_links_recirculation is a boolean. You can do  expect(welsh_article).to_not be_suppress_from_links_recirculation`

spec/models/comfy/cms/page_spec.rb
```diff
@@ -32,6 +84,21 @@
       end
     end
 
+    context 'ignores suppressed articles' do
+      let(:animals_tag) { Tag.create value: 'animals' }
+      let!(:animals_article) { create :english_article, keywords: [animals_tag] }
+
+      it 'order by most matching tags' do
+        create :english_article, keywords: [animals_tag], suppress_from_links_recirculation: true
+        not_suppressed_match_article = create :english_article,
```
tomas-stefano: Can we extract this two line to a let? Or just the first one that is a spec dependency

spec/models/comfy/cms/page_spec.rb
```diff
@@ -150,6 +217,17 @@
...
+    it 'ignores popular articles that have been suppressed' do
+      create :page, page_views: 200, suppress_from_links_recirculation: true
+      create :page, page_views: 1, suppress_from_links_recirculation: false
+      create :page, page_views: 150, suppress_from_links_recirculation: false
+      create :page, page_views: 50, suppress_from_links_recirculation: true
+
+      results = Comfy::Cms::Page.most_popular(3)
+
+      expect(results.map(&:page_views)).to eq([150, 1])
+    end
```
tomas-stefano: Can e add as context and move this create to let as dependency?

spec/controllers/pages_controller_spec.rb
```diff
@@ -137,120 +137,135 @@
...
+        it 'persists page as "draft"' do
+          expect(assigns[:page].state).to eq 'draft'
+        end
       end
-    end
 
-    context 'when passes the "publish" event state' do
-      let!(:page) { create(:page, state: 'draft') }
-      let(:state_event) { 'publish' }
+      context 'when passes the same event state from the page' do
```
tomas-stefano: Good one! :+1:

spec/models/comfy/cms/page_spec.rb
```diff
@@ -1,5 +1,57 @@
...
+        english_article.reload
+
+        expect(english_article.suppress_from_links_recirculation).to be_truthy
+      end
+
+      it 'to not be suppressed' do
+        welsh_article.suppress_from_links_recirculation = false
+        welsh_article.suppress_mirrors_from_links_recirculation
+        english_article.reload
+
```
tomas-stefano: This tree lines are repeating over and over between the specs. I think the logic is shared with a difference between the contexts. Can we move this to a before clause?

app/models/comfy/cms/page.rb
JiggyPete: Unfortunately not

spec/models/comfy/cms/page_spec.rb
JiggyPete: I don't think be_suppress_from_links_recirculation makes sense, specifically 'be_suppress'. If it was be_suppressed_from_links_recirculation I'd be happy with that, however the column name would have to be suppressed_from_links_recirculation which gives it a past tense that I don't think works for the model.

spec/models/comfy/cms/page_spec.rb
JiggyPete: Sure. The intention was to keep the important parts ( the setting of suppress_from_links_recirculation ) local/close to the test for clarity.

spec/models/comfy/cms/page_spec.rb
JiggyPete: In this context the boolean is different and the expectation are different in the two examples.

The previous context is also very similar however we are testing on an english article there, and here we are testing welsh articles.

spec/models/comfy/cms/page_spec.rb
JiggyPete: well spotted

app/models/comfy/cms/page.rb
JiggyPete: done

spec/models/comfy/cms/page_spec.rb
tomas-stefano: Yes, exactly, the boolean is different and the expectation too. Given that:

```ruby
context 'update english article when updating welsh article' do
  before do
    welsh_article.suppress_from_links_recirculation = supressed?
    welsh_article.suppress_mirrors_from_links_recirculation
    english_article.reload
  end
  
  context 'when welsh article supressed' do
    let(:supressed?) { true }
  
    it 'english article should be supressed' do
      expect(english_article.suppress_from_links_recirculation).to be_truthy
    end
  end
  
  context 'when welsh article not supressed' do
    let(:supressed?) { false }
  
    it 'english article should not be supressed' do
      expect(english_article.suppress_from_links_recirculation).to be_falsey
    end
  end
end
```

Something like that (I didn't tested, is just a draft).

tomas-stefano: Btw, If you agreed, the same thing for the inverse above (welsh <-> english). :)

spec/models/comfy/cms/page_spec.rb
tomas-stefano: The past tense could makes sense. Sorry, what did you mean 'you don't think works for the model'? The #suppressed_from_links_recirculation? method after rename the column?

mconnell:
looks good to me :+1:

spec/models/comfy/cms/page_spec.rb
JiggyPete: This looks great!

spec/models/comfy/cms/page_spec.rb
JiggyPete: Yeah, that's what I meant. When calling the getter or setter,
`page.suppress_from_links_recirculation = true/false`
and
`page.suppress_from_links_recirculation?`
`suppress` makes sense.

If we flip that to,
`page.suppressed_from_links_recirculation = true/false`
and
`page.suppressed_from_links_recirculation?`

It reads like (no matter true/false) it has already been suppress'ed' given the past tense.
Especially for the setter.

tomas-stefano: Ok. My english is not the best one to discuss this. I can give a thumbs up on this. But @dougdroper what did you prefer?

dougdroper: Yes, since it is a command and not a question suppress makes more sense. 
Maybe you could have a method on page that asked the question suppressed_from_links_recirculation? but keep the table name / setter as suppress_from_links_recirculation.

tomas-stefano: Ok. Thanks @dougdroper.

PR #139, opened 2015-01-23
Makes cucumber not raises error when running features

tomas-stefano:
* This happen because the rspec is loaded before the cucumber and strange
load error happens

This fix `uninitialized constant RSpec::Matchers::Composable (NameError)`


dougdroper:
:+1:

neoeno:
:+1:

PR #147, opened 2015-01-28
Store MASTalk snippets as escaped strings

ourmaninamsterdam:
And remove Markdown files


dougdroper:
This does lose some readability, but these snippets will hopefully not need to be modified. So I think the improvement in simplicity trumps readability
 :+1:

tomas-stefano:
Agreed with @dougdroper, plus the asset pipeline is too weird/bizarre with requireJS and the markdown files, :\. So, :+1:

PR #149, opened 2015-01-28
Prevents Page.most_popular returning pages with zero views.

neoeno:
To guard against situations where the GA fetch task hasn't been run yet. 

(e.g. when we first deploy to production)

### DIFF
--- a/spec/models/comfy/cms/page_spec.rb
+++ b/spec/models/comfy/cms/page_spec.rb
@@ -234,13 +234,27 @@ RSpec.describe Comfy::Cms::Page do
       expect(results.map(&:page_views)).to eq([200, 152, 100])
     end
 
-    context 'suppressed artictiles' do
+    context 'when articles have zero page views' do
+      before do
+        create :page, page_views: 1, position: 1
+        create :page, page_views: 0, position: 2
+        create :page, page_views: 0, position: 3
+      end
+
+      it 'does not return articles with zero page_views' do
+        results = Comfy::Cms::Page.most_popular(3)
+
+        expect(results.map(&:page_views)).to eq([1])
+      end
+    end
+
+    context 'when there are suppressed articles' do
       let!(:first_suppressed_page) { create :page, page_views: 200, suppress_from_links_recirculation: true }
       let!(:first_not_suppressed_page) { create :page, page_views: 1, suppress_from_links_recirculation: false }
       let!(:second_not_suppressed_page) { create :page, page_views: 150, suppress_from_links_recirculation: false }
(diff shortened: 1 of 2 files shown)


dougdroper:
:+1:

spec/models/comfy/cms/page_spec.rb
```diff
@@ -234,6 +234,16 @@
...
+    it 'should not return articles with zero page_views' do
+      create :page, page_views: 1, position: 1
+      create :page, page_views: 0, position: 2
+      create :page, page_views: 0, position: 3
+
+      results = Comfy::Cms::Page.most_popular(3)
+
+      expect(results.map(&:page_views)).to eq([1])
+    end
+
```
tomas-stefano: Could this spec be a different context? And the spec above there too?

spec/models/comfy/cms/page_spec.rb
```diff
@@ -234,6 +234,16 @@
       expect(results.map(&:page_views)).to eq([200, 152, 100])
     end
 
+    it 'should not return articles with zero page_views' do
```
tomas-stefano: I would rename the 'should not return' to 'does not return'.

spec/models/comfy/cms/page_spec.rb
dougdroper: Why?

spec/models/comfy/cms/page_spec.rb
neoeno: Good catch — testing 101!

spec/models/comfy/cms/page_spec.rb
neoeno: Wasn't sure what you meant here — does the new commit do what you were after?

neoeno:
Squish later.

spec/models/comfy/cms/page_spec.rb
tomas-stefano: @dougdroper I asked because I'm always think in separate contexts when writing the specs. Imho, contexts are a powerful way to make your spec clear and very well organized. And in the long period, contexts keep specs easy to read. Wdyt about it?

tomas-stefano:
@neoeno LGTM after squash :ship:

spec/models/comfy/cms/page_spec.rb
dougdroper: I think contexts are a way to keep contexts together :), and this was in the contexts of the #most_popular scope. Having a context for each 'it' block seems excessive. But saying that I like the way it came out so :+1:

PR #154, opened 2015-02-05
Category manager extension

dougdroper:


app/serializers/page_category_serializer.rb
```diff
@@ -0,0 +1,21 @@
...
+  def title
+    object.label
+  end
+
+  def description
+    object.meta_description
+  end
+
+  def type
+    'guide'
```
tomas-stefano: I don't understand. Why returning 'guide' on the API?

dougdroper: Two fold: 
1) to keep it consistent with the existing API. 
2) the frontend depends on this being here. 

I think you are right in the sense that it is an old name from hippo, that should not really be carried forward. But I think we need to do it in two stages, firstly make sure everything works, then deprecate the category service and then we can remove this, or name it more in line with the domain.

tomas-stefano: Agreed :+1:

app/controllers/api/category_contents_controller.rb
```diff
@@ -0,0 +1,13 @@
+module API
+  class CategoryContentsController < APIController
+    def index
+      @primary_navigation, @secondary_navigation = Comfy::Cms::Category.navigation_categories
+      render json: @primary_navigation, scope: params[:locale].to_sym
```
tomas-stefano: Curiosity. What happens when we pass an inexistent locale?

dougdroper: It would default to Welsh in that case.

tomas-stefano: Is not better validate the locale and returns an error message?

spec/serializers/category_serializer_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+        id: 'test',
+        type: 'category',
+        title: 'en_title',
+        description: 'description_en',
+        parent_id: '',
+        contents: []
+      }
+    end
+
+    let(:child_categories) { nil }
```
tomas-stefano: You can remove this let. The let above is already nil.

app/controllers/api/category_contents_controller.rb
esbel: Is it a public-facing endpoint? If so, please consider white-listing available locales before typecasting user-provided value to `Symbol` objects.

As far as I know, symbols are not garbage collected in major ruby implementations, therefore it leaves space for an unlikely, but possible denial of service attack (assuming the rest of our infrastructure wouldn't prevent that). Some background: http://brakemanscanner.org/docs/warning_types/denial_of_service/

benlovell: 2.2 does symbol GC.

esbel: it's still on 2.1.2, anyway, the point is: try to whitelist user-provided values

benlovell: Move it to 2.2.

tomas-stefano:
:+1:

app/controllers/api/category_contents_controller.rb
dougdroper: Cool, added some validation around the user input, also removed the to_sym since it wasn't really needed. Will also upgrade to 2.2.

esbel: awesome! thanks!

dougdroper:
Ok merging this in now. It is still not live, so will address any potential comments on another PR.

PR #160, opened 2015-02-11
Preview by page type (action plans, news, videos, articles and corporate)

tomas-stefano:
*[DO NOT MERGE]* - Need some work on frontend repository.

PR #161, opened 2015-02-11
Page translation api

tomas-stefano:
[DO NOT MERGE] - Need some work on frontend repository.

### DIFF
--- a/app/models/comfy/cms/page.rb
+++ b/app/models/comfy/cms/page.rb
@@ -5,6 +5,7 @@ class Comfy::Cms::Page < ActiveRecord::Base
   include ComfortableMexicanSofa::IsTaggable::ModelMethods
 
   delegate :identifier, to: :layout, allow_nil: true
+  alias_method :translations, :mirrors
 
   scope :ignore_suppressed, -> { where(suppress_from_links_recirculation: false) }
 
--- a/app/models/comfy/cms/page.rb
+++ b/app/models/comfy/cms/page.rb
@@ -48,6 +48,10 @@ class Comfy::Cms::Page < ActiveRecord::Base
       .reorder('comfy_cms_categorizations.ordinal ASC')
   end
 
+  def translation(locale = :cy)
+    mirrors.find { |page| page.site.locale == locale.to_s }
+  end
+
   def update_page_views(analytics)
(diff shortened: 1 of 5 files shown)


tomas-stefano:
I based this on the branch #160 because I need some code in there. Please review first that PR for me rebase this branch. The only commit that matters here is the 818b53f

dougdroper:
@tomas-stefano, the 'translation' of a page will have the same full_path, only the slugs differ, also both will have the same translation_id. But it doesn't really make your query any better. Comfy::Cms::Page.where(full_path: page.full_path, site: 'cy'), but that is essentially what it is doing under the hood with page.mirrors

tomas-stefano:
Yes, so page.mirrors.first appears the best solution for this case.

dougdroper:
Well you can encapsulate that logic in some helper method: page.translation

tomas-stefano:
Good idea, mate! :)

tomas-stefano:
@dougdroper Done.

asmega:
:+1:

app/models/comfy/cms/page.rb
```diff
@@ -47,6 +48,10 @@ def self.in_category(category_id)
       .reorder('comfy_cms_categorizations.ordinal ASC')
   end
 
+  def translation(locale = :cy)
```
dougdroper: Can this just not be mirrors.first, why do you need to pass the locale to it?

tomas-stefano: To be more agnostic to any language. You can pass any language and will work.

dougdroper: yes, but as it stands you need to know what language you are currently in, so that you can ask for the other, It seems like it is complicating a method on the chance that there may be more languages in the future, or maybe it is just the default I have an issue with. Is there a way to make this work where you can ask a welsh page for the translation without having to explicitly pass :en?

tomas-stefano: Yeah, you alright. I was following the same concept about PageMirror, but is a different context. I will change that.

tomas-stefano: Done

dougdroper: :+1:

dougdroper:
:+1:

PR #166, opened 2015-02-19
Add a fix to avoid select boxes affecting the layout

ourmaninamsterdam:
@moneyadviceservice/frontend

TP#5869

When we have an `<option>` with a long text label it is causing it to interfere with the layout ...
screen shot 2015-02-19 at 10 21 34


benbarnett:
:+1:

tomas-stefano:
:+1:

PR #167, opened 2015-02-19
Add missing fonts to the build.

tomas-stefano:
* We try configure to the asset pipeline in order to see woff and ttf files without any success.
* Force in the build that make sure the fonts are in the public assets folder.

Another solution is appreciated.


ourmaninamsterdam:
:+1:

tomas-stefano:
One more thumbs up? I will merge in some minutes.

ourmaninamsterdam:
:rocket: :ship:

PR #168, opened 2015-02-23
Fix for Meta title input not being populated

ourmaninamsterdam:
Fixed by @tomas-stefano and @ourmaninamsterdam
TP#5890

### DIFF
--- a/app/helpers/pages_helper.rb
+++ b/app/helpers/pages_helper.rb
@@ -1,6 +1,12 @@
 module PagesHelper
-  def dough_component(new_record, components = [])
-    new_record ? { dough_component: components.join(' ') } : {}
+  def page_form_component(page, default: [], optional: [])
+    components = if page.new_record?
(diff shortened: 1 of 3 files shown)


app/helpers/pages_helper.rb
```diff
@@ -1,6 +1,12 @@
 module PagesHelper
-  def dough_component(new_record, components = [])
-    new_record ? { dough_component: components.join(' ') } : {}
+  def page_form_component(page, default: [], optional: [])
+    components = if page.new_record?
```
dougdroper: I don't think I am a fan of assigning variables from conditionals. Can we pull this out into it's own private method. Although annoyingly you have to pass all the variables from this method to another :(, but something like:

{ dough_component: components(page, default, optional).join(' ') }

dougdroper:
:+1:

tomas-stefano:
:+1:

PR #169, opened 2015-02-23
Tool import

dougdroper:


tomas-stefano:
:+1:

spec/serializers/page_category_serializer_spec.rb
```diff
@@ -0,0 +1,32 @@
+describe PageCategorySerializer do
+  let(:layout) { double("Layout", identifier: identifier) }
```
ourmaninamsterdam: `double("Layout", identifier: identifier)`- double quotes.

dougdroper: :+1:

app/serializers/page_category_serializer.rb
```diff
@@ -16,6 +16,10 @@ def description
   end
 
   def type
-    'guide'
+    guide || object.layout.identifier
+  end
+
+  def guide
+    {'article' => 'guide'}[object.layout.identifier]
```
benlovell: Any reason you don't just return a literal here?

dougdroper: What do you mean? 

'article' needs to be transformed into 'guide' to keep backwards compatibility with the old API, since that is what hippo called it, and all the other layouts translate as is. I am sure we can remove this silliness once we retire the old category manager.

benlovell: Exactly, just return `'guide'` as there is no other option here, well, other than if you index into this hash with a non-existing key from `object.layout.identifier`. Or am I missing something obvious?

benlovell: I'm a twerp.

benlovell:
:+1:

PR #173, opened 2015-02-27
Category locale

dougdroper:


app/controllers/pages_controller.rb
```diff
@@ -40,6 +40,6 @@ def build_file
   end
 
   def set_categories
-    @categories = Comfy::Cms::CategoriesListPresenter.new(@site.categories.of_type('Comfy::Cms::Page'))
+    @categories = Comfy::Cms::CategoriesListPresenter.new(Comfy::Cms::Category.where(site: Comfy::Cms::Site.first))
```
tomas-stefano: Can we move the english site to a parent controller and shared here?

app/controllers/categories_controller.rb
```diff
@@ -57,4 +60,8 @@ def category_params
   def find_category
     @category = Comfy::Cms::Category.find(params[:id])
   end
+
+  def english_site
+    @english_site ||= Comfy::Cms::Site.first
```
tomas-stefano: I know that the english is the first row in the database, but can we make find_by locale here? (locale or path or label) Imho, this leads to a more consistent logic that don't rely on the rows order.

tomas-stefano:
:+1:

asmega:
:+1:

PR #179, opened 2015-03-10
Increase width of Page title column

ourmaninamsterdam:
Also stops all columns from narrowing too much.


dougdroper:
:+1:

ourmaninamsterdam:
:+1: Pseudo thumb from @tomas-stefano at my desk.

PR #181, opened 2015-03-10
Fix external link suffix issue

ourmaninamsterdam:
It wasn't inserting the external link suffix when the last piece of text in a sentence was selected.


tomas-stefano:
:+1:

PR #182, opened 2015-03-10
Simple version - Hippo report tool

tomas-stefano:
This is the simple version of the hippo diff page between hippo and Contento :tm: 

The perfect solution will be scraping the content xml file from Hippo instead expecting on the  tmp dir. Will do this in a separate PR.

### DIFF
--- a/app/controllers/hippo/diffs_controller.rb
+++ b/app/controllers/hippo/diffs_controller.rb
@@ -1,0 +1,21 @@
+class Hippo::DiffsController < Comfy::Admin::Cms::BaseController
+  before_action :check_admin
+
+  def show
+    @hippo_diff = Cms::HippoDiff.new(data: data)
+  end
+
+  private
+
+  def data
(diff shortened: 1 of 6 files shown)


app/controllers/hippo/diffs_controller.rb
```diff
@@ -0,0 +1,21 @@
...
+  before_action :check_admin
+
+  def show
+    @hippo_diff = Cms::HippoDiff.new(data: data)
+  end
+
+  private
+
+  def data
+    File.read(Rails.root.join('tmp', filename))
```
tomas-stefano: Again, I will write in case you didn't read the comment on the PR. The perfect solution will be scraping the content xml file on the Hippo CMS System, instead expecting on the tmp dir. Will do this in a separate PR.

dougdroper:
:+1:

Tr4pSt3R:
:+1: Looks good to meet

PR #185, opened 2015-03-18
Use require_dependency for comfy

tomas-stefano:
Use require_dependency in order to stop errors in development mode, when rails reload the code.


ourmaninamsterdam:
:+1:

PR #190, opened 2015-03-23
Fix URL formatter

ourmaninamsterdam:
When editing the beginning of a URL in the URL field, the cursor would jump to the end of the field. This stops that from happening by keeping the cursor in the editing position.

TP#6072


dougdroper:
:+1:

tomas-stefano:
:+1:

PR #193, opened 2015-03-24
Add unsaved changes prompt

ourmaninamsterdam:
Adds a component which watches elements within a container for any user input on form fields and `contentEditable` elements within. If the user makes a change and tries to navigate away or reload the page, a confirmation prompt is shown:

screen shot 2015-03-24 at 16 33 12


dougdroper:
very nice :+1: also like that it works on browser close :)

tomas-stefano:
:heart: :heart: :+1: :+1:

PR #199, opened 2015-04-08
Allow placing of cursor in URL field

ourmaninamsterdam:
Fixes a bug where users were unable to set the cursor position within the URL field.

@moneyadviceservice/team-bee 

id:6214


Guntrisoft:
:+1:

PR #201, opened 2015-04-15
Hippo export tweaks

asmega:
- fixes hippo pdf export
- add timeout and exception handling to rackspace upload

### DIFF
--- a/lib/hippo_file.rb
+++ b/lib/hippo_file.rb
@@ -4,7 +4,7 @@ class HippoFile < HippoXmlParser::Document
   end
 
   def blob
-    binary_blob = (find_property('hippo:text') || find_property('jcr:data')).value
+    binary_blob = (find_property('jcr:data') || find_property('hippo:text')).value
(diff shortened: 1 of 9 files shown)


lib/hippo_file.rb
```diff
@@ -4,7 +4,7 @@ def filename
   end
 
   def blob
-    binary_blob = (find_property('hippo:text') || find_property('jcr:data')).value
+    binary_blob = (find_property('jcr:data') || find_property('hippo:text')).value
```
mconnell: Wondering if it's worth having a test that shows the precedence for the hippo blob?

mconnell:
Other than the inline comment :+1:

darokel:
:+1:

lib/hippo_file.rb
asmega: There wasn't an original spec for this and I've not added one. I've only tested by actually running the script.

PR #203, opened 2015-04-17
add Page#publishable?

asmega:
by default a page is publishable unless tagged with do-not-publish
when not publishable admin UI to publish a page is removed

this feature will be used to provide content to engines
where we do not want articles to be published but still available
this is done by using the preview functionality as a back door

### DIFF
--- a/app/models/comfy/cms/page.rb
+++ b/app/models/comfy/cms/page.rb
@@ -82,4 +82,8 @@ class Comfy::Cms::Page < ActiveRecord::Base
   def ever_been_published?
     revisions.map {|revision| revision.data[:event]}.include?('published')
   end
+
+  def publishable?
+    keywords.all?(&:publishable?)
+  end
 end
--- a/spec/models/comfy/cms/page_spec.rb
+++ b/spec/models/comfy/cms/page_spec.rb
@@ -1,5 +1,4 @@
 RSpec.describe Comfy::Cms::Page do
-
   describe '#suppress_from_links_recirculation' do
     let(:english_site) { create :site, is_mirrored: true }
     let(:welsh_site) { create :site, :welsh, is_mirrored: true }
@@ -278,4 +277,17 @@ RSpec.describe Comfy::Cms::Page do
     end
   end
 
+  describe '#publishable?' do
+    context 'when tagged with do-not-publish' do
+      it 'returns false' do
+        allow(subject).to receive(:keywords) { [Tag.new(value: 'do-not-publish')] }
+        expect(subject).to_not be_publishable
+      end
+    end
+
+    it 'returns true' do
+      allow(subject).to receive(:keywords) { [Tag.new(value: 'a-tag')] }
+      expect(subject).to be_publishable
+    end
+  end
 end
--- a/app/models/tag.rb
+++ b/app/models/tag.rb
@@ -20,4 +20,8 @@ class Tag < ActiveRecord::Base
   def in_use?
     taggings.exists?
   end
+
+  def publishable?
+    value != 'do-not-publish'
+  end
 end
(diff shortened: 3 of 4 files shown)


app/models/comfy/cms/page.rb
```diff
@@ -82,4 +82,8 @@ def suppress_mirrors_from_links_recirculation
   def ever_been_published?
     revisions.map {|revision| revision.data[:event]}.include?('published')
   end
+
+  def publishable?
+    keywords.map{ |tag| tag.value }.exclude?('do-not-publish')
```
benlovell: `keywords.none? { |t| t.value == 'do-not-publish' }`

spec/models/comfy/cms/page_spec.rb
```diff
@@ -278,4 +277,17 @@
     end
   end
 
+  describe '#publishable?' do
+    context 'when tagged with do-not-publish' do
+      it 'returns false' do
+        allow(subject).to receive(:keywords) { [Tag.new(value: 'do-not-publish')] }
+        expect(subject.publishable?).to be_falsey
```
benlovell: `expect(subject).to_not be_publishable` is nicer perhaps?

spec/models/comfy/cms/page_spec.rb
```diff
@@ -278,4 +277,17 @@
...
+      end
+    end
+
+    it 'returns true' do
+      allow(subject).to receive(:keywords) { [Tag.new(value: 'a-tag')] }
+      expect(subject.publishable?).to be_truthy
```
benlovell: `expect(subject).to be_publishable` is nicer perhaps?

app/models/comfy/cms/page.rb
benlovell: For bonus points you could push that check into `Tag` behind a predicate `publishable?` (although you'd need to flip the logic from `none?` to `any?` of course. If it has real meaning, it probably belongs over there in `Tag`.

spec/models/comfy/cms/page_spec.rb
asmega: agreed

spec/models/comfy/cms/page_spec.rb
asmega: yep

app/models/comfy/cms/page.rb
asmega: no idea `none?` even exisited!

asmega: are you thinking Tag.publishable?(article_or_thing) # true or false

benlovell: `Tag#publishable?` that does the check for `value != 'do-not-publish` or perhaps the inverse `Tag#embargoed?` that does the check for `value == 'do-not-publish'`

asmega: i'd prefer to keep the word publish somewhere. so `unpublishable` or `not_publishable` or something alike as opposed to introduce addtional terminology

asmega: not sure if that's better or worse

benlovell: I agree, but there are two concerns here. The `#publishable?` thing has meaning for `Page`, which will still be around, it just essentially delegates its logic to `Tag#(publishable?|embargoed?)` whichever which way you decide to implement it.

Perhaps:

``` ruby
# Page#publishable?
def publishable?
  tags.all?(&:publishable?)
end

# Tag#publishable?
def publishable?
  value == 'blah-what-ever'
end
```

asmega: yep looks good. A Monday task I think

asmega:
looks about right now

benlovell:
Beautiful :heart_eyes: :ship:

PR #206, opened 2015-05-08
Add componentNames to JS modules

ourmaninamsterdam:
Pending changes in Dough use `componentName` to stamp the attribute on the component element when it is initialised https://github.com/moneyadviceservice/dough/pull/219


darokel:
:+1:

PR #207, opened 2015-05-11
CMS api now serves HTML

asmega:
- At the moment the CMS generates markdown content which is then processed by the client.
- This changes the API so it is now the job of the CMS to generate HTML.
- I think this is the job of the CMS and not the client hence the move
- This will save every client from having implement something to process the markdown (with custom snippets) to HTML

### DIFF
--- a/features/step_definitions/edit_page_steps.rb
+++ b/features/step_definitions/edit_page_steps.rb
@@ -9,7 +9,7 @@ When(/^I preview the article$/) do
 end
 
 Then(/^I should see the Draft Article$/) do
-  expect(JSON.parse(preview_page.text)['blocks'].find {|b| b["identifier"] == 'content'}['content']).to eq('test')
+  expect(JSON.parse(preview_page.text)['blocks'].find {|b| b["identifier"] == 'content'}['content']).to include('test')
 end
 
 Then(/^I should not be able to see live draft article$/) do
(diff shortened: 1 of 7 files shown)


spec/models/block_composer_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+      let(:blocks) { [] }
+
+      it 'is an empty string' do
+        expect(composer.to_html).to eql("\n")
+      end
+    end
+
+    context 'non-content block provided' do
+      let(:non_content_block) do
+        { 'identifier' => 'not_content', 'content' => 'Definitely not a Content Block' }
```
tomas-stefano: Yeah. Definitely! :)

tomas-stefano:
:+1: :+1:

features/step_definitions/edit_page_steps.rb
```diff
@@ -9,7 +9,7 @@
 end
 
 Then(/^I should see the Draft Article$/) do
-  expect(JSON.parse(preview_page.text)['blocks'].find {|b| b["identifier"] == 'content'}['content']).to eq('test')
+  expect(JSON.parse(preview_page.text)['blocks'].find {|b| b["identifier"] == 'content'}['content']).to include('test')
```
tomas-stefano: This assertion is too smart. I know that was already there. I would move this to a less smart assertion.

spec/models/block_composer_spec.rb
asmega: I just borrowed this from frontend. So not my code. I'll delete from frontend once this is integrated.

features/step_definitions/edit_page_steps.rb
asmega: Not gonna touch this at the moment.

spec/models/block_composer_spec.rb
tomas-stefano: I don't have any problem with this line. For me it's fine. It just a funny and a bad comment.

features/step_definitions/edit_page_steps.rb
tomas-stefano: Ok. For me it's fine this PR. :+1:

PR #213, opened 2015-05-14
make full_path in serializer useful

asmega:
this use to return the slug we now return /locale/page_type/slug
although this is a breaking change to the API I am not aware of any external
apps using they full_path at the moment

cc @tomas-stefano


tomas-stefano:
:+1: Since we only have the frontend as a client (and now the partners with the feed) I will not be concern about any external apps that uses full_path node. Given that :+1: :ship:

PR #214, opened 2015-05-14
Seed categories correctly

darokel:
Updates the seeds file to set up categories needed to run Contento locally correctly.


asmega:
:+1:

PR #215, opened 2015-05-14
add published_at to serializer

asmega:
cc @tomas-stefano

### DIFF
--- a/app/models/comfy/cms/page.rb
+++ b/app/models/comfy/cms/page.rb
@@ -90,4 +90,14 @@ class Comfy::Cms::Page < ActiveRecord::Base
   def publishable?
     keywords.all?(&:publishable?)
   end
+
+  def published_at
+    if published?
+      updated_at
+    elsif scheduled?
+      scheduled_on
+    elsif ever_been_published?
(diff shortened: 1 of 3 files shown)


app/models/comfy/cms/page.rb
```diff
@@ -90,4 +90,14 @@ def ever_been_published?
   def publishable?
     keywords.all?(&:publishable?)
   end
+
+  def published_at
+    if published?
+      updated_at
+    elsif scheduled?
+      scheduled_on
```
tomas-stefano: I am curious. Is this scheduled expected as a business rule/acceptance criteria?

asmega: Yes. Cultivate asked for it and it's in the ticket 5116

tomas-stefano: Cool. Thanks

asmega: After re-reading you comment. The requirement is for a `published_at` timestamp. This would be covered if an article has been scheduled

darokel:
:+1: :ship:

PR #217, opened 2015-05-18
fix page#published_at when event is nil

asmega:
this may by due to off data in production
this bug manifests when the first revision event is nil

### DIFF
--- a/app/models/comfy/cms/page.rb
+++ b/app/models/comfy/cms/page.rb
@@ -97,7 +97,7 @@ class Comfy::Cms::Page < ActiveRecord::Base
     elsif scheduled?
       scheduled_on
     elsif ever_been_published?
-      revisions.detect { |revision| Array(revision.data[:event]).include?('published') }.created_at
+      revisions.detect { |revision| revision.data[:event] == 'published' }.created_at
     end
   end
 end
--- a/app/models/comfy/cms/page.rb
+++ b/app/models/comfy/cms/page.rb
@@ -97,7 +97,7 @@ class Comfy::Cms::Page < ActiveRecord::Base
     elsif scheduled?
       scheduled_on
     elsif ever_been_published?
-      revisions.detect { |revision| revision.data[:event].include?('published') }.created_at
+      revisions.detect { |revision| Array(revision.data[:event]).include?('published') }.created_at
     end
   end
 end
--- a/spec/models/comfy/cms/page_spec.rb
+++ b/spec/models/comfy/cms/page_spec.rb
@@ -313,10 +313,24 @@ RSpec.describe Comfy::Cms::Page do
     context 'been unpublished' do
       subject { create(:page, state: 'draft') }
 
-      before { subject.revisions.create!(data: {event: 'published'}, created_at: 3.hours.ago) }
+      context 'been published once' do
+        before { subject.revisions.create!(data: {event: 'published'}, created_at: 3.hours.ago) }
 
-      it 'returns last published revision datetime' do
-        expect(subject.published_at).to be_within(10.seconds).of(3.hours.ago)
+        it 'returns last published revision datetime' do
+          expect(subject.published_at).to be_within(10.seconds).of(3.hours.ago)
+        end
+      end
+
+      # possibly due to bad data in production
+      context 'when data event is nil' do
+        before do
+          subject.revisions.create!(data: {event: nil}, created_at: 7.hours.ago)
+          subject.revisions.create!(data: {event: 'published'}, created_at: 8.hours.ago)
+        end
+
+        it 'returns last published revision datetime' do
+          expect(subject.published_at).to be_within(10.seconds).of(8.hours.ago)
+        end
       end
     end
 


spec/models/comfy/cms/page_spec.rb
```diff
@@ -313,10 +313,24 @@
...
+
+      # possibly due to bad data in production
+      context 'when data event is nil' do
+        before do
+          subject.revisions.create!(data: {event: nil}, created_at: 7.hours.ago)
+          subject.revisions.create!(data: {event: 'published'}, created_at: 8.hours.ago)
+        end
+
+        it 'returns last published revision datetime' do
+          expect(subject.published_at).to be_within(10.seconds).of(8.hours.ago)
```
urfolomeus: Would it be worth using Timecop (or similar) here?

misaka:
:+1:

spec/models/comfy/cms/page_spec.rb
asmega: I've not used timecop before. The docs seem to suggest one would use it for more complicated scenarios. How would you use it here?

urfolomeus: Mainly so that I knew the exact value and wasn't relying on the time generated being within 10 seconds. :)

I know it's highly unlikely that this limit will be breached but it could introduce heisenbugs further down the line.

app/models/comfy/cms/page.rb
```diff
@@ -97,7 +97,7 @@ def published_at
     elsif scheduled?
       scheduled_on
     elsif ever_been_published?
-      revisions.detect { |revision| revision.data[:event].include?('published') }.created_at
+      revisions.detect { |revision| Array(revision.data[:event]).include?('published') }.created_at
```
urfolomeus: It's maybe just me but I find this a little confusing to parse when reading. Would `revision.data[:event].try(:include?, 'published')` be more readable?

spec/models/comfy/cms/page_spec.rb
asmega: with `Timecop.freeze`

```
     Failure/Error: expect(subject.published_at).to eql(3.hours.ago)

       expected: Mon, 18 May 2015 08:50:41 BST +01:00
            got: Mon, 18 May 2015 08:50:41 BST +01:00
```

urfolomeus: :D I'll get me coat

asmega: Not sure why that's happened. I just expected it to work

app/models/comfy/cms/page.rb
misaka: :+1: ... I had to think about the `Array()` bit a little too, this probably would've communicated it better.

spec/models/comfy/cms/page_spec.rb
paulanthonywilson: I agree about preferring TimeCop to fix specs at certain times, but it seems a bit out of scope for this PR.

app/models/comfy/cms/page.rb
asmega: I'd rather array-ify it so the next thing knows exactly what API it should use and opposed to a nil guard

spec/models/comfy/cms/page_spec.rb
urfolomeus: I've seen this issue before, but normally where timezones are involved. The pattern I tend to use is:

```
now = Time.zone.now
Timecop.freeze(now)
subject = Thing.create({data: {event: 'published'}})
expect(subject.published_at).to eql(now)
Timecop.return
```

urfolomeus: Probably not worth introducing just for this case though. Was just asking in general. Sorry for the yak shave! :)

asmega:
```
    before :each do
      Timecop.freeze(Time.zone.now)
    end

    after :each do
      Timecop.return
    end
```

Is what I used and it didn't work

app/models/comfy/cms/page.rb
urfolomeus: Perhaps pull it out to a named method then to make it a bit more obvious (to dummies like me at least ;) ) what the intention is?

spec/models/comfy/cms/page_spec.rb
urfolomeus: I think @paulanthonywilson is right. Timecop is probably better looked at at a higher level. Apologies for the tangent :)

app/models/comfy/cms/page.rb
paulanthonywilson: Are we not switching from ```String#include?``` to ```Array#include?``` here?

misaka: Don't think so. `#try?` should be calling `include?` on the object it's being called on, `revision.date[:event]` here. In any case, @asmega is probably right and this would just force this code to be bigger than it needs to be, considering the return value of `detect` is being used. Nice idea, though. :)

paulanthonywilson: Yeah, but revision.data[:event] returns a String (or apparently nil), at least according to the specs. This is turning it into an array of strings, containing one string, or an empty array.

voila_capture 2015-05-18_12-15-22_pm

misaka: Good catch. Sounds like the `Array()` fix could potentially also change the behaviour depending on what values we can get for `revision.date[:event]`.

asmega: Done. I'm now checking against a string

tomas-stefano:
:+1:

misaka:
Just a friendly reminder to add a test against `published_but_being_edited`. :)

urfolomeus:
Looks good, thanks :) :+1:

spec/models/comfy/cms/page_spec.rb
urfolomeus: So it turns out, after some digging, that `.be-within` is recommended for this sort of match now due to comparing floating point numbers. Apologies for the detour.

Every day is a school day! :)

PR #220, opened 2015-06-08
Persist published at

asmega:
- Persist `published_at` when a page is published
- We will need to retrofit the `published_at` field for pages that are already published or have been published at some point

### DIFF
--- a/app/models/comfy/cms/page.rb
+++ b/app/models/comfy/cms/page.rb
@@ -90,14 +90,7 @@ class Comfy::Cms::Page < ActiveRecord::Base
   def publishable?
     keywords.all?(&:publishable?)
   end
-
-  def published_at
-    if published?
-      updated_at
-    elsif scheduled?
-      scheduled_on
-    elsif ever_been_published?
-      revisions.detect { |revision| revision.data[:event] == 'published' }.created_at
-    end
-  end
 end
+
+Comfy::Cms::Page.state_machine.events[:publish].timestamp = :published_at
+Comfy::Cms::Page.state_machine.events[:publish_changes].timestamp = :published_at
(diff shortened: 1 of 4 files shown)


urfolomeus:
LGTM :) :+1:

app/models/comfy/cms/page.rb
```diff
@@ -90,14 +90,7 @@ def ever_been_published?
...
-    elsif scheduled?
-      scheduled_on
-    elsif ever_been_published?
-      revisions.detect { |revision| revision.data[:event] == 'published' }.created_at
-    end
-  end
 end
+
+Comfy::Cms::Page.state_machine.events[:publish].timestamp = :published_at
+Comfy::Cms::Page.state_machine.events[:publish_changes].timestamp = :published_at
```
tomas-stefano: Why? Maybe I misunderstood the commit message.

app/models/comfy/cms/page.rb
```diff
@@ -90,14 +90,7 @@ def ever_been_published?
...
-
-  def published_at
-    if published?
-      updated_at
-    elsif scheduled?
-      scheduled_on
-    elsif ever_been_published?
-      revisions.detect { |revision| revision.data[:event] == 'published' }.created_at
-    end
-  end
```
tomas-stefano: This was the new code to feeds works as expected, right?

asmega: This data is now persisted to the database so we can pull it out from there

app/models/comfy/cms/page.rb
asmega: So when an article is published we set the published_at field using the state machine. Instead of doing the code which attempting to dynamically work out the published_at value

tomas-stefano:
LGTM :+1: (Don't forget to test publishing, publishing while editing, turn in draft, and publish after, and after the recirculation drivers script too)

asmega:
@tomas-stefano I've added an additional test so when the record is updated and save (which affects the timestamps) it does not affect the `published_at` field

tomas-stefano:
Cool. :+1: :+1:

PR #222, opened 2015-06-17
page links now consider their layout

asmega:
previously all pages were incorrectly considered articles only


tomas-stefano:
:+1: :+1:

PR #225, opened 2015-06-26
Add category images

asmega:
users can associate a small and large image to a category
via the category page

### DIFF
--- a/app/serializers/category_serializer.rb
+++ b/app/serializers/category_serializer.rb
@@ -60,8 +60,6 @@ class CategorySerializer < ActiveModel::Serializer
     object.third_level_navigation
   end
 
-  private
-
   def has_small_image?
     object.small_image
   end
--- a/app/serializers/category_serializer.rb
+++ b/app/serializers/category_serializer.rb
@@ -25,14 +25,6 @@ class CategorySerializer < ActiveModel::Serializer
     end
   end
 
-  def has_small_image?
-    object.small_image
-  end
-
-  def has_large_image?
-    object.large_image
-  end
-
   def contents
     (
       object.child_categories <<
@@ -67,4 +59,15 @@ class CategorySerializer < ActiveModel::Serializer
   def third_level_navigation
     object.third_level_navigation
   end
+
+  private
+
+  def has_small_image?
+    object.small_image
+  end
+
+  def has_large_image?
+    object.large_image
+  end
+
 end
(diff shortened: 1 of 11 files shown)


Tr4pSt3R:
I run the spec and this is one of the errors I am getting: 
```
pry(#<RSpec::ExampleGroups::CategorySerializer::English>)> subject.as_json
ArgumentError: bad argument (expected URI object or URI string)
from /opt/rubies/2.1.2/lib/ruby/2.1.0/uri/common.rb:537:in `convert_to_uri'
```

spec - code:
```
it 'returns the contents of the category' do
  expect(subject.as_json).to eq(expected_en)
end
```

Could you run your specs and see if this is the case for you too please @asmega?

app/serializers/category_serializer.rb
```diff
@@ -1,10 +1,38 @@
...
+
+  def small_image_url
+    if has_small_image?
+      URI.join(ActionController::Base.asset_host, object.small_image.try(:file).try(:url).to_s).to_s
+    end
+  end
+
+  def large_image_url
+    if has_large_image?
+      URI.join(ActionController::Base.asset_host, object.large_image.try(:file).try(:url).to_s).to_s
```
tomas-stefano: Can we move this to a private method?

tomas-stefano: By the way, Is Code Climate suppose to catch these changes. Is it running for Contento/CMS? cc @darokel

config/environments/production.rb
```diff
@@ -55,7 +55,7 @@
   # config.cache_store = :mem_cache_store
 
   # Enable serving of images, stylesheets, and JavaScripts from an asset server.
-  # config.action_controller.asset_host = "http://assets.example.com"
+  config.action_controller.asset_host = "http://comfy.moneyadviceservice.org.uk"
```
tomas-stefano: Double quotes. Is this not better on a ENV var?

asmega: why would this be better as an ENV var?

tomas-stefano: This config is served by production, but dev/test is diff. If something that could differ from environment it is better to move to a configuration management or an ENV var You already have https://github.com/moneyadviceservice/cms/blob/master/.env.dev where you can use those variables.

asmega:
@Tr4pSt3R my fault. now fixed

app/serializers/category_serializer.rb
asmega: done

config/environments/production.rb
asmega: I've swapped the quotes and add the missing test environment config. I'm not sure how using .env.dev will work. There is no corresponding .env file on QA nor are the all the environment variables found that file present in configuration management (scripts repo)

darokel:
Just one comment on approach but apart from that the code is :+1:

PR #227, opened 2015-06-29
Category promo

asmega:
- This allows the cms to add promoted items to a category

### DIFF
--- a/app/models/category_promo.rb
+++ b/app/models/category_promo.rb
@@ -1,0 +1,5 @@
+class CategoryPromo < ActiveRecord::Base
+  PROMO_TYPES = %w{ blog recommended popular calculator tool }
+
+  belongs_to :category, class: Comfy::Cms::Category
+end
--- a/app/models/category_promo.rb
+++ b/app/models/category_promo.rb
@@ -2,4 +2,8 @@ class CategoryPromo < ActiveRecord::Base
   PROMO_TYPES = %w{ blog recommended popular calculator tool }
 
   belongs_to :category, class: Comfy::Cms::Category
+
+  validates :title, presence: true
+  validates :promo_type, inclusion: PROMO_TYPES
+  validates :locale, inclusion: %w{ en cy }
 end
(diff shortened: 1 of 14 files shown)


spec/models/category_promo_spec.rb
```diff
@@ -0,0 +1,5 @@
+describe CategoryPromo do
+  it 'associations' do
```
neoeno: we've got this in sentence format [here](https://github.com/moneyadviceservice/cms/pull/227/files#diff-5cfe7e4f7bab0f6559c519ec3aa95b18R29), but you could equally use `specify 'associations' do`

app/models/category_promo.rb
```diff
@@ -0,0 +1,5 @@
+class CategoryPromo < ActiveRecord::Base
+  PROMO_TYPES = %w{ blog recommended popular calculator tool }
+
+  belongs_to :category, class: Comfy::Cms::Category
+end
```
urfolomeus: I know your protecting against submissions with no title but shouldn't that be backed up with validations in here?

Same for ensuring promo type is included in PROMO_TYPES.

neoeno:
:+1: from me, though @darokel makes a good point

spec/models/category_promo_spec.rb
neoeno: Or `it { is_expected.to belong_to(:category) }`

app/models/category_promo.rb
urfolomeus: Also for locale?

spec/models/category_promo_spec.rb
asmega: The latter option looks good

app/models/category_promo.rb
asmega: Agreed. I've now added these validations

tomas-stefano:
:+1:

darokel:
:+1: :ship:

urfolomeus:
:+1:

PR #228, opened 2015-07-08
post process html output to wrap table elements

asmega:
- Kramdown library isn't very extensisble so not done at that level
- this is done to enable scrollable tables


benbarnett:
LGTM :+1:

tomas-stefano:
LGTM :+1:

PR #229, opened 2015-07-14
inserting an image adds description as alt text

asmega:
- Also changed file description label to 'Alt text' so editors are aware of what the field will be used for


jonalport:
:+1:

mconnell:
:+1:

PR #230, opened 2015-07-15
Fix create page errors

asmega:
- On failed page creation this no longer raises an unexpected exception due to missing variables
- Slight hack to populate flash has with something meaningful. Comfy doesn't provide any hooks to make this possible.


rnalexander:
:+1:

darokel:
:+1: Though I think it would be better to redefine the comfy controller that's needed and override a particular method if we have to refactor the original controller actions. Polymorphism and method overriding allows this :smile:

asmega:
@darokel you'd essentially have to copy/paste the existing method then modify. Ideally we don't want to modify the comfy code, I know that we have already diverged but we don't want to make it harder for ourselves later.

PR #232, opened 2015-07-22
add basic create and view redirects functionality

asmega:


benbarnett:
:+1:

darokel:
I don't think we should have a mix of haml AND erb in the code base, it'll just be confusing.

app/controllers/redirects_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+
+  def index
+    @redirects = Redirect.all
+  end
+
+  private
+
+  def load_admin_site
+    @site = Comfy::Cms::Site.first
+  end
```
tomas-stefano: Why we have the site?

spec/models/redirect_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+    end
+
+    it { is_expected.to validate_presence_of(:source)  }
+    it { is_expected.to validate_presence_of(:destination)  }
+    it { is_expected.to validate_presence_of(:redirect_type)  }
+    it { is_expected.to validate_uniqueness_of(:source) }
+
+    it 'validates source and destination not identical' do
+      subject.destination = 'foo'
+      expect(subject).to_not be_valid
```
tomas-stefano: A personal preference, I always prefer to be specific on this specs. For two reasons, it is very difficult to trace where is the problem. If you change the implementation in some way, RSpec won't produce a descriptive error message on the record be_valid. It will simply says "expected valid? to return false, got true." 
Did you try:

``ruby
  expect(subject).to have(1).error_on(:destination)
```

And change the implementation to add the error:

```ruby
  errors.add(:destination, :source_duplicated)
```

And in the locale you create a key called source_duplicated.

```
subject.errors[:source]

Comment (tomas-stefano) [inline spec/models/redirect_spec.rb]: What do you think about it?

File: app/controllers/redirects_controller.rb (hunk 1 above)
Comment (asmega) [inline app/controllers/redirects_controller.rb]: I've now removed this. Comfy overrides render. As we're inheriting from `Comfy::Admin::Cms::BaseController`, there was code using `site` if the template could not be found. now that the templates exist (which they didn't before the tests were create) this is no longer needed

Comment (tomas-stefano) [inline app/controllers/redirects_controller.rb]: Ok :+1:

File: spec/models/redirect_spec.rb (hunk 2 above)
Comment (asmega) [inline spec/models/redirect_spec.rb]: I've changed the error from base to destination attribute as I think that its better.

Regarding the specific error I think the test description is succinct enough and will be displayed should the test fail.

**Comment by asmega** [comment]
@darokel I prefer erb but can convert to haml if needed. I think it'll be less confusing as there will be less haml!

File: spec/controllers/redirects_controller_spec.rb [hunk 3]
```diff
@@ -0,0 +1,63 @@
+describe RedirectsController do
+  describe 'auth' do
+    context 'when not logged in' do
+      it 'redirects user to login' do
+        get :new
+        expect(response).to_not be_success
```
urfolomeus: Would it not be better to say `expect(response).to redirect_to(login_path)` (or whatever the url helper is for the login path)?

spec/controllers/redirects_controller_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+  before do
+    create(:site)
+    sign_in current_user
+  end
+
+  describe '#new' do
+    it 'works' do
+      get :new
+      expect(response).to be_success
+    end
```
urfolomeus: Likewise, this doesn't really test what this action is doing. There's no test to assert that the assign is happening or that the new template is being rendered.

spec/controllers/redirects_controller_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+        expect(flash[:danger]).to be_present
+      end
+    end
+  end
+
+  describe '#index' do
+    it 'works' do
+      get :index
+      expect(response).to be_success
+    end
```
urfolomeus: Ditto above.

spec/controllers/redirects_controller_spec.rb
asmega: done

urfolomeus: Sweet :)

urfolomeus:
Apart from a few comments, a big ole :+1:

spec/controllers/redirects_controller_spec.rb
asmega: done

spec/controllers/redirects_controller_spec.rb
asmega: done

urfolomeus:
:+1:

spec/models/redirect_spec.rb
tomas-stefano: Ok :+1:

PR #236, opened 2015-07-31
redirect source must start with a slash validation

asmega:


spec/models/redirect_spec.rb
```diff
@@ -28,15 +26,27 @@
     end
 
     describe 'format validation' do
-      it 'must begin with /en or /cy' do
-        subject.destination = '/en/foo'
-        expect(subject).to be_valid
+      context 'for source' do
+        it 'must begin /' do
```
Tr4pSt3R: ```it 'must begin with /' do```

asmega: done

Tr4pSt3R:
:100: :+1:

PR #239, opened 2015-08-04
add additional namespaced api endpoint

asmega:
to prevent collision with the rest of the application

### DIFF
--- a/spec/requests/api_spec.rb
+++ b/spec/requests/api_spec.rb
@@ -1,0 +1,21 @@
+RSpec.describe "old api endpoint", type: :request do
+  let!(:site) { create(:site, path: 'en', locale: 'en', is_mirrored: true) }
+  let!(:article_layout) { create(:layout, :article, site: site, identifier: 'article') }
+  let!(:article) { create(:page, site: site, layout: article_layout) }
+
+  it "returns the article at /en/articles/article.json" do
+    get "/en/articles/#{article.slug}.json"
+    expect(JSON.parse(response.body)['full_path']).to eql("/en/articles/#{article.slug}")
+  end
+end
+
+RSpec.describe "new api endpoint", type: :request do
+  let!(:site) { create(:site, path: 'en', locale: 'en', is_mirrored: true) }
+  let!(:article_layout) { create(:layout, :article, site: site, identifier: 'article') }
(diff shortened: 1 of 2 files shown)


benbarnett:
Well steered :+1:

jonalport:
:+1: :shipit:

spec/requests/api_spec.rb
```diff
@@ -0,0 +1,21 @@
+RSpec.describe "old api endpoint", :type => :request do
+  let!(:site) { create(:site, path: 'en', locale: 'en', is_mirrored: true) }
+  let!(:article_layout) { create(:layout, :article, site: site, identifier: 'article') }
+  let!(:article) { create(:page, site: site, layout: article_layout) }
+
+  it "returns the article at /en/articles/article.json" do
+    get "/en/articles/#{article.slug}.json"
+    expect(JSON.parse(response.body)['full_path']).to eql("/en/articles/#{article.slug}")
```
darokel: If there are going to be more scenarios where you need to parse the JSON response, I recommend extracting this into a support method.

urfolomeus:
:1:

spec/requests/api_spec.rb
urfolomeus: Would you not extract that into a helper method _when_ you have more scenarios, rather than in anticipation of more scenarios?

asmega: This request spec as a quick sanity check to ensure the new and old endpoints continue to work.

The detailed controller spec is still available at https://github.com/moneyadviceservice/cms/blob/master/spec/controllers/api/content_controller_spec.rb

darokel: Now or later, either way lines will be shaved.

spec/requests/api_spec.rb
```diff
@@ -0,0 +1,21 @@
+RSpec.describe "old api endpoint", :type => :request do
```
tomas-stefano: Hash rocket

asmega: fixed

tomas-stefano:
:+1:

spec/requests/api_spec.rb
asmega: extracting this to a method will involve adding more lines.

PR #240, opened 2015-08-05
Allow for unusual characters within anchor links

jonalport:


benbarnett:
:+1:

asmega:
this will have to marry up with the ruby implementation for auto ids otherwise we'll get dead links

asmega:
which can be found here https://github.com/gettalong/kramdown/blob/master/lib/kramdown/converter/base.rb#L232

jonalport:
@asmega: This is a valid point. The implementations are the same, except that the ruby allows for multiple consecutive hyphens, which is suboptimal IMHO.

Example string: "Hello , there"

In ruby: "hello--there"

In JS: "hello-there"

Can we edit the ruby or not so much?

asmega:
I don't fancy patching as it's a third party lib so will probably leave as is.

jonalport:
Thought as much :)

benbarnett:
post-25179-you-can-do-it-gif-waterboy-img-jc2v

jonalport:
Haha! Ok, I've got this.

jonalport:
@asmega: it is done

asmega:
:+1:

PR #241, opened 2015-08-06
Set up New Relic

darokel:
New Relic has already been set up for Contento within Puppet but the yaml config needs to be temporary left in this repo as we don't update Contento through Puppet for production.


urfolomeus:
:+1:

PR #245, opened 2015-08-14
Add new relic yaml for production

darokel:


config/newrelic.yml.erb
```diff
@@ -0,0 +1,41 @@
...
+    monitor_mode: false
+
+  staging:
+    <<: *default_settings
+    app_name: Contento (Staging)
+
+  production:
+    <<: *default_settings
+    app_name: Contento (Production)
+<% end %>
```
munckymagik: Ok so to check I've understood this right:

* No configuration is set for QA env
* Configuration is set for development, test, staging and production
* Staging and production are the same (except for the app_name)
* Test mode reports nothing to New Relic (I don't know the diff between this and QA env?)
* Development does report but also enables extra developer features

Am I on the right lines?

darokel: This PR needs to be closed. Essentially, we've set up new relic for all the key environments within Puppet which outputs the correct yaml file. The CMS was originally configured via Puppet but not finished so Puppet does not deploy it the same way as the other projects. We have to manually deploy the CMS to production.

This is changing for the new Azure environment. The CMS has been set up correctly there. As a temporary fix we've "hot" pushed a seperate new relic yaml file for the CMS production environments as that's the only one that Puppet can't create since it doesn't deploy it.

Here are some further PRs for reference https://github.com/moneyadviceservice/scripts/pulls?utf8=%E2%9C%93&q=is%3Apr+is%3Aclosed+new+relic+

PR #249, opened 2015-08-20
Feature toggling

asmega:
- Since our infrastructure isn't automated, I'm not able to inject things at environmental level. So I've done this at app level.
- Detect the current environment via various means
- Pull in the appropriate feature toggles


darokel:
The list of features will eventually have to be moved out of here and into Puppet once we move to Azure. As Contento is automated there.

asmega:
it doesn't have to be as it should be portable but yes it should be moved to the environment.

darokel:
:+1:

PR #250, opened 2015-08-21
add validation to prevent redirect loops and hops

asmega:


tomas-stefano:
:+1:

app/models/redirect.rb
```diff
@@ -10,6 +10,8 @@ class Redirect < ActiveRecord::Base
   validates :destination, presence: true, format: { with: /\A\/en|\/cy/ }
   validates :redirect_type, presence: true, inclusion: { in: REDIRECT_TYPES }
   validate  :validate_different_source_and_destination
+  validate  :destination_matches_existing_source
```
paulanthonywilson: Am I reading the logic right? Shouldn't these be

```
validate :destination_does_not_match_existing_source
validate :source_does_not_match_existing_destination
```

asmega: yep, i'll tweak that now

paulanthonywilson: Cool, sorry for spotting it late.

PR #253, opened 2015-08-24
redirects can be soft deleted

asmega:


munckymagik:
:+1: Looks good. I'm assuming a way to re-activate is done elsewhere? Or are we just wanting to give ourselves the option of restoring deleted redirects?

asmega:
@munckymagik The plan was for end users not to be able reactivate. They are able to add a new redirect to replace the old one if needs be. keeping it as soft delete means we can keep an audit trail

munckymagik:
Cool ok. Understood. Looks good.

PR #261, opened 2015-09-16
Use public url for comfy if available.

misaka:
In our current environment the same hostname is being used for both internal and external access to the CMS, by resolving the hostname to an internal or external IP depending on where it's being resolved from. To fix this, we need to allow the CMS to have a separate public hostname in the environment config.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -56,7 +56,11 @@ Rails.application.configure do
   # config.cache_store = :mem_cache_store
 
   # Enable serving of images, stylesheets, and JavaScripts from an asset server.
-  config.action_controller.asset_host = ENV['MAS_CMS_URL']
+  config.action_controller.asset_host = if ENV.has_key?('MAS_CMS_PUBLIC_URL')
+                                          ENV['MAS_CMS_PUBLIC_URL']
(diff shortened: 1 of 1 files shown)


misaka:
Ya, I don't mind. I tend to lean away from the ternary operator in general and in this case it would take it over the 80-characters-wide limit I usually follow. But if it falls in better with MAS's code style I can switch it.

darokel:
I guess the first is always more readable so maybe go with that and wrap `if ENV.has_key? 'MAS_CMS_PUBLIC_URL'` with parens `if ENV.has_key?('MAS_CMS_PUBLIC_URL')`

Tr4pSt3R:
it's ripe for a ternary one-liner :)
One-line it and it will be all good to go

misaka:
Amended to add parens to `has_key?`'s arguments.

@Tr4pSt3R I still think this is more readable, as @darokel mentioned also, especially with the line being a bit longer.

darokel:
:+1:

paulanthonywilson:
:+1: 

(I think it's more readable as it is than with a ternary, but these things can be a matter of taste)

PR #262, opened 2015-09-17
Make tests of file listings in admin page not fail.

misaka:


asmega:
:+1:

misaka:
@asmega I've changed this so that it sorts the list of files in the expectation instead of when it generates the list of files. It seems to make sense to move the sorting closer to the expectation.

Tr4pSt3R:
I wonder if it would be better if ```filenames_of(type: type).sort``` was spelt out as to filetypes are being expected?

misaka:
@Tr4pSt3R I think I've just made that change, actually. Could you refresh and double-check that that's what you meant?

Tr4pSt3R:
I feel it will be clearer to spell out the filenames of a definite type. i.e.

```expect(shown_filenames).to eq(filenames_of(type: type).sort)``` should be
```expect(shown_filenames).to eq[‘file_a’, ‘file_b]```
the eq part, I feel should always be explicit

Tr4pSt3R:
:+1: good to go though

asmega:
:+1:

PR #267, opened 2015-09-24
Add redirects to nav menu

asmega:
- only admins can see this in their nav
- only admins will be able to access the redirects functionality


Tr4pSt3R:
:+1:

PR #269, opened 2015-10-01
redirects support aspx, pdf and html extensions

asmega:


benbarnett:
aspx?! I won't ask

:+1:

rnalexander:
:+1: :shipit:

PR #271, opened 2015-10-01
link inserter uses correct url

asmega:
all links inserted were previously hard coded to the page type of
articles. this now corretly looks up the page type for the link


rnalexander:
:+1: :shipit:

PR #273, opened 2015-10-05
redirects with parameters are considered as dupes

asmega:
this prevents hops eg.

/en/origin => /en/middle?param=a
/en/middle => /en/final

the above scenario above is not considered valid
as `middle` is considered a duplicate


jonalport:
:+1: Cunning

PR #277, opened 2015-10-12
[DEV APPROVED]add validations for redirect source extensions

asmega:


rnalexander:
:+1:

benbarnett:
:+1:

asmega:
@darokel yes it's just the same thing. nested if statements are nested if statements. i'll change if you can find a regexp that fulfils all three if statements

PR #286, opened 2015-11-02
[DEV APPROVED] Category expand

johnplayer1982:
Allows a user to expand the list of categories on the CMS homepage

screen shot 2015-11-02 at 08 56 44

### DIFF
--- a/app/assets/javascripts/comfortable_mexican_sofa/admin/modules/CategoryExpand.js
+++ b/app/assets/javascripts/comfortable_mexican_sofa/admin/modules/CategoryExpand.js
@@ -1,0 +1,140 @@
+define(['jquery','DoughBaseComponent'], function ($, DoughBaseComponent) {
+'use strict';
+
+  var defaultConfig = {
+    categoriesCharacterDelimiter: ','
+  },
+  CategoryExpandProto,
+  $cells;
+
+  function CategoryExpand($el, config) {
+
+    CategoryExpand.baseConstructor.call(this, $el, config, defaultConfig);
+
+  }
+
+  DoughBaseComponent.extend(CategoryExpand);
+
+  CategoryExpand.componentName = 'CategoryExpand';
+
+  CategoryExpandProto = CategoryExpand.prototype;
+
+  CategoryExpandProto.init = function(initialised) {
+
+    this._initialiseVariables();
+    this._initialiseCells();
+    this._addAjaxListeners();
+    this._addUIEventListeners();
+
+    this._initialisedSuccess(initialised);
+    return this;
+
+  };
+
+  CategoryExpandProto._initialiseVariables = function() {
+
+    $cells = this.$el.find('.categories');
+
+  };
+
+  CategoryExpandProto._addAjaxListeners = function() {
+
+    $('#filter_form').on('ajax:complete', $.proxy( this._ajaxCompleteHandler, this));
+    $('#filter_form').on('ajax:beforeSend', $.proxy( this._ajaxBeforeSendHandler, this));
+
+  };
+
+  CategoryExpandProto._addUIEventListeners = function() {
+
+    $cells.on('click', this._cellsClickHandler);
+
+  };
+
+  CategoryExpandProto._removeUIEventListeners = function() {
+
+    $cells.off('click', this._cellsClickHandler);
+
+  };
+
+  CategoryExpandProto._initialiseCells = function() {
+
+    var self = this;
+
+    $cells.each(function() {
+
+      var $currentCell = $(this),
+      cellText = $currentCell.text(),
+      categoryArray = cellText.split(self.config.categoriesCharacterDelimiter),
+      categoryCount = categoryArray.length,
+      cellData = {
+        '$currentCell' : $currentCell,
+        'cellText' : cellText,
+        'categoryArray' : categoryArray,
+        'categoryCount' : categoryCount
+      };
+
+      if (cellData.categoryCount > 1) {
+        self._addExpanders(cellData);
+      };
+
+    });
+
+  };
+
+  CategoryExpandProto._addExpanders = function(cellData) {
+
+    var delim = this.config.categoriesCharacterDelimiter;
+    var firstCategory = cellData.categoryArray.splice(0, 1)[0];
+    var numberOtherCategories = cellData.categoryCount - 1;
+    var otherCategoriesText = cellData.categoryArray.join(delim);
+
+    cellData.$currentCell.html(
+          firstCategory +
+          ' <div><a href="#" class="more">+' + numberOtherCategories + ' more</a>' +
+          '<span style="display:none;" class="allCategories">'+ otherCategoriesText +
+          ' <a href="#" class="less">Hide</a></span></div>'
+        );
+
+  };
+
+  CategoryExpandProto._ajaxBeforeSendHandler = function(evt) {
+
+    this._removeUIEventListeners();
+
+  };
+
+  CategoryExpandProto._ajaxCompleteHandler = function(evt) {
+
+      this._initialiseVariables();
+      this._initialiseCells();
+      this._addUIEventListeners();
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/comfortable_mexican_sofa/admin/modules/CategoryExpand.js
```diff
@@ -0,0 +1,71 @@
...
+      this.addExpanders();
+    }, this));
+
+    this._initialisedSuccess(initialised);
+      return this;
+    };
+
+    CategoryExpandProto.addExpanders = function(){
+
+    var self = this;
```
benbarnett: This indentation is a little out

app/assets/javascripts/comfortable_mexican_sofa/admin/modules/CategoryExpand.js
```diff
@@ -0,0 +1,71 @@
...
+
+    $('#filter_form').on('ajax:complete', $.proxy(function() {
+      this.addExpanders();
+    }, this));
+
+    this._initialisedSuccess(initialised);
+      return this;
+    };
+
+    CategoryExpandProto.addExpanders = function(){
```
benbarnett: Should have spaces after this so functions are written like this `function() {`

app/assets/javascripts/comfortable_mexican_sofa/admin/modules/CategoryExpand.js
```diff
@@ -0,0 +1,71 @@
...
+
+  CategoryExpand.componentName = 'CategoryExpand';
+
+  CategoryExpandProto = CategoryExpand.prototype;
+
+  CategoryExpandProto.init = function(initialised) {
+
+    this.addExpanders();
+
+    $('#filter_form').on('ajax:complete', $.proxy(function() {
```
benbarnett: Can you move these 3 lines into another method called `this.addListeners()`?

app/assets/javascripts/comfortable_mexican_sofa/admin/modules/CategoryExpand.js
```diff
@@ -0,0 +1,71 @@
...
+
+    this._initialisedSuccess(initialised);
+      return this;
+    };
+
+    CategoryExpandProto.addExpanders = function(){
+
+    var self = this;
+
+    $('.js-filter-pages td:nth-child(3)').addClass('categories');
```
benbarnett: Did we speak to @asmega about getting the classes added directly?

app/assets/javascripts/comfortable_mexican_sofa/admin/modules/CategoryExpand.js
```diff
@@ -0,0 +1,71 @@
...
+
+    this.$el.find('.more').click(function(event){
+      event.preventDefault();
+      $(this).hide().prev().hide();
+      $(this).next().show();
+    });
+
+    this.$el.find('.less').click(function(event){
+      event.preventDefault();
+      $(this).parent().hide().prev().show().prev().show();
```
benbarnett: This seems a little fragile on first reading... do you think there's a way we could tighten it up so we know we're targetting the right thing?

app/assets/javascripts/comfortable_mexican_sofa/admin/modules/CategoryExpand.js
asmega: @johnplayer1982 can do that here https://github.com/moneyadviceservice/cms/blob/master/app/views/comfy/cms/pages/_page.html.haml

johnplayer1982: Got it thanks Phil

app/assets/javascripts/comfortable_mexican_sofa/admin/modules/CategoryExpand.js
```diff
@@ -0,0 +1,72 @@
...
+
+  CategoryExpand.componentName = 'CategoryExpand';
+
+  CategoryExpandProto = CategoryExpand.prototype;
+
+  CategoryExpandProto.init = function(initialised) {
+
+    this.addExpanders();
+
+    this.addListeners = function() {
```
benbarnett: Sorry this isn't quite right. I meant to make it a new method like `CategoryExpandProto.addExpanders = function() {` is, and then call it from here with `this.addListeners()`. Does that make sense?

app/assets/javascripts/comfortable_mexican_sofa/admin/modules/CategoryExpand.js
```diff
@@ -0,0 +1,76 @@
+define(['jquery','DoughBaseComponent'], function ($, DoughBaseComponent) {
+'use strict';
+
+  var defaultConfig = {
+    categoriesCharacterCount: 20
+  };
+  var CategoryExpandProto;
```
nmiddleweek: Small thing but it would be a good idea to combine these into one var statement...

```
var defaultConfig = {
       categoriesCharacterCount: 20
   },
   CategoryExpandProto;
```

app/assets/javascripts/comfortable_mexican_sofa/admin/modules/CategoryExpand.js
```diff
@@ -0,0 +1,76 @@
...
+
+  function CategoryExpand($el, config) {
+    CategoryExpand.baseConstructor.call(this, $el, config, defaultConfig);
+  }
+
+  DoughBaseComponent.extend(CategoryExpand);
+
+  CategoryExpand.componentName = 'CategoryExpand';
+
+  CategoryExpandProto = CategoryExpand.prototype;
```
nmiddleweek: If this was done to reduce typing/ readability, good idea... An alternative to this is shown below although I've not seen it in this code-base yet...

```
CategoryExpand.prototype = {
   'init' = function() {
    },
   'aMethod' = function() {
    },
   'anotherMethod' = function() {
    }
}
```

This is just an FYI...

app/assets/javascripts/comfortable_mexican_sofa/admin/modules/CategoryExpand.js
```diff
@@ -0,0 +1,76 @@
...
+          categoryCount = categories.split(',').length - 1,
+          categoriesCharacterCount = self.config.categoriesCharacterCount;
+
+      if (categories.indexOf(',') >= 1) {
+        $(this).html(
+          categories.slice(0,categoriesCharacterCount)+
+          '<a href="#" class="more"> +' + categoryCount + ' see all</a>'+
+          '<span style="display:none;" class="allCategories">'+ categories.slice(categoriesCharacterCount ,categories.length)+' <a href="#" class="less">Hide</a></span>'
+        );
+      };
```
nmiddleweek: It might be a nice idea to pull this out as a JS $template partial using say mustache, handlebars or lodash (whichever we have) and then we can then pass it parameters to more easily build up the categories list.

We can go through this tomorrow if you like.

app/assets/javascripts/comfortable_mexican_sofa/admin/modules/CategoryExpand.js
```diff
@@ -0,0 +1,76 @@
...
+      $(this).hide();
+      $(this).parent().find('.allCategories').show();
+    });
+
+    this.$el.find('.less').click(function(event) {
+      event.preventDefault();
+      $(this).parent().hide();
+      $(this).closest('.categories').find('.more').show();
+    });
+
```
nmiddleweek: It would be nice to have these bound to $more and $less variables and/or use delegation to handle these click events. It would also be good to remove the .parent() references however in some situations it can be useful. Perhaps we can create a $categories variable as well.

asmega:
:+1:

nmiddleweek:
:+1:

PR #288, opened 2015-11-10
Bugsnag

darokel:
Integrate Bugsnag error reporting. API key set in Puppet https://github.com/moneyadviceservice/scripts/pull/450


asmega:
:+1:

PR #296, opened 2015-12-07
[WIP] - use pandoc not openoffice to parse word docs

asmega:
also makes uploading word docs partially usable compared to not usable
at all

unpend tests around word doc uploads

remove word to markdown gem as no longer needed


tomas-stefano:
:+1: :+1:

PR #300, opened 2016-02-03
[DEV APPROVED] Roles

KateWilkinson:
This is some ground work in preparation for adding a new 'External Editor' user to the application. We originally considered a separate `Role` model, but realised the following leads to a much simpler implementation.

* Adds `#role` to `Comfy::Cms::User` for helping with authorisation. The existing behaviour for regular and admin users has not been changed.
* Updates the user form to have a dropdown instead of a checkbox
* Data migration to move add the correct role to existing users
* Cleanup migration to remove the admin boolean from users table

### DIFF
--- a/db/migrate/20160203102911_migrate_users_to_have_role.rb
+++ b/db/migrate/20160203102911_migrate_users_to_have_role.rb
@@ -1,0 +1,23 @@
+class MigrateUsersToHaveRole < ActiveRecord::Migration
+  def up
+    Comfy::Cms::User.all.each do |user|
+      user.update_attribute(:role, new_role_for(user))
+    end
+  end
+
+  def down
+    Comfy::Cms::User.all.each do |user|
+      user.update_attribute(:admin, user.admin?)
+    end
+  end
+
+  private
+
+  def new_role_for(user)
+    if user.read_attribute(:admin)
+      Comfy::Cms::User.roles[:admin]
+    else
+      Comfy::Cms::User.roles[:author]
+    end
+  end
+end
--- a/db/migrate/20160203102911_migrate_users_to_have_role.rb
+++ b/db/migrate/20160203102911_migrate_users_to_have_role.rb
@@ -17,7 +17,7 @@ class MigrateUsersToHaveRole < ActiveRecord::Migration
     if user.read_attribute(:admin)
       Comfy::Cms::User.roles[:admin]
     else
-      Comfy::Cms::User.roles[:author]
+      Comfy::Cms::User.roles[:user]
     end
   end
 end
--- a/features/step_definitions/user_management_steps.rb
+++ b/features/step_definitions/user_management_steps.rb
@@ -46,7 +46,11 @@ Then(/^the new user should not be able to log in$/) do
 end
 
 Given(/^I am not an admin user$/) do
-  @current_user = Comfy::Cms::User.create!(admin: false, email: 'test@tester.com', password: 'password')
+  @current_user = Comfy::Cms::User.create!(
+    role: Comfy::Cms::User.roles[:author],
+    email: 'test@tester.com',
+    password: 'password'
+  )
 end
 
 Then(/^I should not be able to visit the user management page$/) do
--- a/features/step_definitions/user_management_steps.rb
+++ b/features/step_definitions/user_management_steps.rb
@@ -47,7 +47,7 @@ end
 
 Given(/^I am not an admin user$/) do
   @current_user = Comfy::Cms::User.create!(
-    role: Comfy::Cms::User.roles[:author],
+    role: Comfy::Cms::User.roles[:user],
     email: 'test@tester.com',
     password: 'password'
   )
--- a/spec/factories/users.rb
+++ b/spec/factories/users.rb
@@ -4,6 +4,6 @@ FactoryGirl.define do
     name     { Faker::Name.name }
     password '12345678'
     password_confirmation '12345678'
-    role 1
+    role 'admin'
   end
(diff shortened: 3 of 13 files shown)


spec/models/comfy/cms/user_spec.rb
```diff
@@ -1,4 +1,20 @@
 describe Comfy::Cms::User do
+  describe 'validations' do
+    describe 'role' do
+      it { should validate_presence_of(:role) }
+
+      it 'allows the role to be set as a "user"' do
```
tomas-stefano: I would move this two íts to another describe block, since is not about validations.

KateWilkinson:
@tomas-stefano thanks for spotting that! We've adjusted as suggested.

db/migrate/20160203102911_migrate_users_to_have_role.rb
```diff
@@ -0,0 +1,23 @@
...
+  private
+
+  def new_role_for(user)
+    if user.read_attribute(:admin)
+      Comfy::Cms::User.roles[:admin]
+    else
+      Comfy::Cms::User.roles[:user]
+    end
+  end
+end
```
tomas-stefano: I would move this to a rake task and delete this migration and the next one (removal of admin field). ˆThen after 'everybody' is happy and the data is migrated, you can delete the admin field in another deploy, but doing in steps. But  if you think that do at once will work fine, then is ok. Wdyt?

features/step_definitions/user_management_steps.rb
```diff
@@ -46,7 +46,11 @@
 end
 
 Given(/^I am not an admin user$/) do
-  @current_user = Comfy::Cms::User.create!(admin: false, email: 'test@tester.com', password: 'password')
+  @current_user = Comfy::Cms::User.create!(
```
Tr4pSt3R: Hi @KateWilkinson :), this can be easily replaced with a FactoryGirl for 'user'. 

See here: https://github.com/moneyadviceservice/cms/blob/master/spec/factories/users.rb

db/migrate/20160203102911_migrate_users_to_have_role.rb
mconnell: We could move it out to a separate task, but the reason for it being part of this is to prevent a misunderstanding in behaviour.

The effect of using `enum role` with an `'admin'` value, is that the `Comfy::Cms::User` obtains an `#admin?` method, which led to a little bit of confusion when an `admin` attribute is left lying around on the model, as it also dynamically creates an `#admin?` method. 

In our scenario, the `#admin?` method for our new role functionality appears to be the last method loaded and therefore it is its functionality we would be using. However, without digging further, I couldn't be confident in saying that this method would _always_ be the last one loaded. And it also means we have duplicate data lingering within the application.

In this case, my preference would be to keep things as is, but we can change it if you think we should?

features/step_definitions/user_management_steps.rb
mconnell: Hi @Tr4pSt3R. The change was simply to allow a test to continue to pass. I feel it's not within the scope of this particular PR to refactor the implementation of the test.

If I switched it to using factory girl the only thing that would want to change would be `Comfy::Cms::User.create!` would become `FactoryGirl.create(:user,`. As it's a feature step definition, I would expect the email address/password to potentially be re-used by other steps.

However, it does raise a good point about reviewing the current codebase and seeing if there are things we can maybe make a pass thru and tidy up wholesale.

mconnell: @Tr4pSt3R we've had to add an additional step for the next part of work, so we're going to address this change to a factory while we're touching that file.

Tr4pSt3R: Hey Mark, that sounds good :)

db/migrate/20160203102911_migrate_users_to_have_role.rb
tomas-stefano: Ok, I said what I said because of the risks of changing data with difficult to rollback if something go wrong. But since the CMS users data is small, and in this case I would say that it is better do what you feel that is good to do it. I will give the thumbs, for it and leave you guys decide what is best.

tomas-stefano:
:+1: :+1:  Good work.

spec/factories/users.rb
```diff
@@ -4,6 +4,6 @@
     name     { Faker::Name.name }
     password '12345678'
     password_confirmation '12345678'
-    admin true
+    role 1
```
munckymagik: I don't understand why we are using integer literals to refer to the role names.

My understanding of the Rails `enum` type is that it hides the underlying numeric field type used in the database and  instead lets us use meaningful/readable named values within the code. E.g. http://www.justinweiss.com/articles/creating-easy-readable-attributes-with-activerecord-enums/

Can this and the other literals not be changed to `role 'admin' or `subject.role = 'admin'`? That is far easier to understand from a maintainer/reviewer perspective.

If we use integer literals in our code then we've completely lost all the benefits we could have had.

Maybe I've misunderstood something, but sorry this seems confusing.

munckymagik: Sorry if this comes across as "arsey". I shouldn't write comments in the morning :)

This is a great piece of work, just this one thing change would make it more great IMO.

KateWilkinson:
@munckymagik thanks for your advice. Latest commit removes those references.

munckymagik:
:+1: great.

tomas-stefano:
Don't forget to delete the branch (unless you're working on it) ;)

PR #301, opened 2016-02-05
Add editor

KateWilkinson:
Using the groundwork laid in PR #300, a new editor role has been added.

The existing behaviour stating only admin users have access to the admin section has not been changed, therefore an editor does not have access to the user management page and only an admin can create a new editor user profile.


tomas-stefano:
:+1: :+1:

cazrin:
:+1:

PR #303, opened 2016-02-08
[DEV APPROVED] Limit editor functionality

KateWilkinson:
Previously we introduced the notion of an 'Editor' role into the cms. This PR is a continuation of that work to enforce the permitted abilities for an Editor.
 
* Removes ability to create a new page if current user is an editor
* Removes ability to schedule, publish, unpublish if current user is an editor
* Updated `PageRegister#save` to restrict editor capability.

### DIFF
--- a/app/helpers/pages_helper.rb
+++ b/app/helpers/pages_helper.rb
@@ -38,6 +38,11 @@ module PagesHelper
     end
   end
 
+  def display_additional_button_menu?(page, user)
+    return false if user.editor? || page_state_buttons(page).empty?
+    page.publishable?
+  end
+
   private
 
   def preview_domain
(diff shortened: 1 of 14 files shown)


app/helpers/pages_helper.rb
```diff
@@ -38,6 +38,12 @@ def render_form_for_layout(site, page)
     end
   end
 
+  def display_additional_button_menu?(page, user)
+    return false if user.editor?
+    return false if page_state_buttons(page.current_state).empty?
```
tomas-stefano: Can we use a || on this two conditionals? Then we don't need two returns.

KateWilkinson:
@tomas-stefano thanks - we've adjusted as you suggested.

tomas-stefano:
:+1: :+1:

paulanthonywilson:
:+1:

PR #304, opened 2016-02-10
[DEV APPROVED] Editors changing content should trigger a notification email

KateWilkinson:
This PR adds functionality for a notification email to be triggered whenever a user with an 'editor' role saves changes to content. We've added mailjet and reuse existing env vars from scripts.

### DIFF
--- a/app/mailers/revision_mailer.rb
+++ b/app/mailers/revision_mailer.rb
@@ -1,0 +1,12 @@
+class RevisionMailer < ActionMailer::Base
+  TO_ADDRESS = 'welsh.editors@moneyadviceservice.org.uk'
+  SUBJECT    = 'Content updated by External Editor'
+
+  default from: ENV['MAILJET_DEFAULT_FROM_ADDRESS']
+
+  def external_editor_change(user:, page:)
+    @page = page
+    @user = user
+    mail(to: TO_ADDRESS, subject: SUBJECT)
+  end
+end
--- a/config/initializers/mailjet.rb
+++ b/config/initializers/mailjet.rb
@@ -1,0 +1,5 @@
+Mailjet.configure do |config|
+ config.api_key = ENV['MAILJET_API_KEY']
+ config.secret_key = ENV['MAILJET_SECRET_KEY']
+ config.default_from = ENV['MAILJET_DEFAULT_FROM_ADDRESS']
+end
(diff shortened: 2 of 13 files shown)


app/mailers/revision_mailer.rb
```diff
@@ -0,0 +1,9 @@
+class RevisionMailer < ActionMailer::Base
+  default from: ENV['MAILJET_DEFAULT_FROM_ADDRESS']
+
+  def external_editor_change(user:, page:)
+    @page = page
+    @user = user
+    mail(to: 'welsh.editors@moneyadviceservice.org.uk', subject: 'Content updated by External Editor')
```
tomas-stefano: I would move this strings to constants.

KateWilkinson: @tomas-stefano we've moved these to constants, thanks.

app/mailers/revision_mailer.rb
```diff
@@ -0,0 +1,12 @@
+class RevisionMailer < ActionMailer::Base
+  TO_ADDRESS = 'welsh.editors@moneyadviceservice.org.uk'
+  SUBJECT    = 'Content updated by External Editor'
+
+  default from: ENV['MAILJET_DEFAULT_FROM_ADDRESS']
```
cazrin: I don't know what the convention around environment variables is in the project, but I learned a handy tip recently that helps identify errors. If you use `ENV.fetch('key')` it will raise an error saying the key doesn't exist if you forget to set it.

It's slightly longer than the above, which will just return `nil` if the key doesn't exist, but it helps when you're trying to debug why it isn't working and it's because you forgot to set an environment variable.

config/initializers/mailjet.rb
```diff
@@ -0,0 +1,5 @@
+Mailjet.configure do |config|
+ config.api_key = ENV["MAILJET_API_KEY"]
```
cazrin: Is Rubocop not complaining about these evil double quotes?

cazrin:
:+1: Just had a couple of comments that you can ignore if you want :D

app/mailers/revision_mailer.rb
tomas-stefano: If the `fetch` is used and the environment variable is not set, the application will not start due to the fetch error. If the `[]`  is used then I can assume that only the mailer will have the bug. So, we can conclude, that It is a matter of choice on what effect could the worst in case the environment variable is not set.

tomas-stefano:
:+1:

PR #308, opened 2016-02-17
Add home page seed

KateWilkinson:
We've been looking at adding the ability to edit additional content on the frontend website. In the process of this, we had to run frontend locally against CMS locally. During this process, we encountered an issue where we were unable to load the website as there was no "Home Page" within contento.

Upon further investigation, we realised that in order for the frontpage of frontend to render, we needed to create a 'home_page' layout, and create a `Page` within contento with the slug 'the-money-advice-service', which uses this layout.

This PR adds to `seeds.rb` an additional snippet to generate the 'home_page' layout, and a very basic published homepage.

We were struggling with coming up with good names for the classes we've wrapped the behaviour in, also if anyone has any suggestions on how to better structure the code, that would be awesome!

### DIFF
--- a/spec/lib/cms/layout_builder_spec.rb
+++ b/spec/lib/cms/layout_builder_spec.rb
@@ -1,0 +1,92 @@
+RSpec.describe Cms::LayoutBuilder do
+  let!(:english_site) { create(:site, is_mirrored: true) }
+  let!(:welsh_site)   { create(:site, :welsh, is_mirrored: true) }
+
+  describe '.add_home_page!' do
+    before { Cms::LayoutBuilder.add_home_page! }
+
+    context 'English site' do
+      let(:layout) { english_site.layouts.reload.first }
+
+      it 'creates a layout for the English site' do
+        expect(english_site.layouts.count).to eq(1)
+      end
+
+      it 'sets the label for the layout to be "Home Page"' do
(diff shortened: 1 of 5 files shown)


Tr4pSt3R:
It looks good @KateWilkinson. Something that was definitely need sooner rather than later. I can't see anything wrong with it :) 
:+1:

spec/lib/cms/layout_builder_spec.rb
```diff
@@ -0,0 +1,92 @@
+RSpec.describe Cms::LayoutBuilder do
+  let!(:english_site) { FactoryGirl.create(:site, is_mirrored: true) }
+  let!(:welsh_site)   { FactoryGirl.create(:site, :welsh, is_mirrored: true) }
```
tomas-stefano: The factory girl is included in the specs and it using only the create call as default. I would remove the namespace to follow the pattern. Wdyt?

tomas-stefano: The factory girl is included in the specs and it using only the create call as default. I would remove the namespace to follow the pattern. Wdyt?

tomas-stefano:
:+1: :+1:

KateWilkinson:
Thanks @Tr4pSt3R and @tomas-stefano !

PR #309, opened 2016-02-17
editors should be able to create drafts from an existing published page

KateWilkinson:
Users with an editor role, are able to make changes to pages, but not publish them.

We discovered whilst looking at QA test data, that we had missed one of the possible events (when a editor makes a draft version of an already published page).

This small PR addresses this issue.


tomas-stefano:
:+1:

paulanthonywilson:
:+1:

PR #310, opened 2016-02-17
Force an article to exist before creating a homepage

KateWilkinson:
Now that we've updated the seeds.rb to generate a homepage, we've discovered an unexpected side effect...

Comfy weirdly defaults to the first layout used, which means if you've seeded a db from scratch, it now decides that the home_page layout should be the default, see gif:

This behaviour *only* happens when seeding a database from scratch.

This PR adds an example article into the database before adding the homepage, so it mimics the behaviour you would expect to see locally, or on a server with existing data like QA.


tomas-stefano:
:+1:

Tr4pSt3R:
:+1:

PR #311, opened 2016-02-18
Use label instead of meta_title, as latter is not guaranteed to exist

KateWilkinson:
Small PR. Switch email to use @page.label instead of @page.meta_title.

When an external editor saved changes to an English page, a link to the page was sent via email, however if changes were made to a Welsh page the link was not visible in the email as meta titles are not automatically copied across. Labels will always exist in both pages, so the label is a more appropriate attribute to use for the link in the email.


spencer:
:+1:

tomas-stefano:
:+1:

PR #314, opened 2016-02-18
Editors can ALSO save_changes_as_draft

KateWilkinson:
There is an additional state where an Editor making changes to a page, should be able to save changes as a new draft.

To recap why we have 4 states here now:

* save_unsaved – Saving a new page which has been generated automatically, but no content exists.
* save_changes – We have a saved page which is unpublished, and changes are being saved
* save_changes_as_draft – There is a published page, and the user is saving a draft page.
* save_draft_changes – After creating a draft from a published page, the user can then save further changes to that draft.

### DIFF
--- a/app/models/page_register.rb
+++ b/app/models/page_register.rb
@@ -74,7 +74,7 @@ class PageRegister
   end
 
   def editor_can_perform_event?
-    %w(save_changes save_unsaved save_draft_changes).include? state_event
+    %w(save_changes save_unsaved save_draft_changes save_changes_as_draft).include? state_event
   end
 
   def update_state_if_new_page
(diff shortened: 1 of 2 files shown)


app/models/page_register.rb
```diff
@@ -74,7 +74,7 @@ def ensure_permission_to_save!
   end
 
   def editor_can_perform_event?
-    %w(save_changes save_unsaved save_draft_changes).include? state_event
+    %w(save_changes save_unsaved save_draft_changes save_changes_as_draft).include? state_event
```
tomas-stefano: Can you move to a constant, please?

jongilbraith:
:+1: 

Just noticed this myself before lunch as it happens.

app/models/page_register.rb
KateWilkinson: @tomas-stefano done! thanks

cazrin:
:+1:

Tr4pSt3R:
:+1:

PR #315, opened 2016-02-18
editors are allowed to alter scheduled pages

KateWilkinson:
Currently, editors are only able to save pages which are not published or scheduled. This change enables editors to save changes to pages that are already scheduled.


tomas-stefano:
:+1:

Tr4pSt3R:
:+1:

urfolomeus:
:+1:

PR #317, opened 2016-02-19
6981 cms upload through azure

habitullence:


tomas-stefano:
Is there a reason to rollback the .env?

tomas-stefano:
:+1:

.env
```diff
@@ -1,14 +1,21 @@
...
+AZURE_ASSETS_STORAGE_CMS_ACCOUNT_CONTAINER='container'
+AZURE_ASSETS_STORAGE_CMS_ACCOUNT_KEY='key'
+AZURE_ASSETS_STORAGE_CMS_ACCOUNT_NAME='name'
+FARADAY_HOST='localhost:5000'
+FARADAY_X_FORWARDED_PROTO='http'
 FRONTEND_URL="localhost:5000"
 GA_API_EMAIL_ADDRESS="google.service.account@example.com"
 GA_PRIVATE_KEY_PATH="path/to/private.key"
+MAILJET_DEFAULT_FROM_ADDRESS=notmonitored@notify.moneyadviceservice.org.uk
+MAS_CMS_URL="http://localhost:3000/"
```
Tr4pSt3R: Hey @habitullence is this intentional?

misaka:
@tomas-stefano Yes, it turns out that the tests in Go have a problem if the `.env` is not there. :) We could come up with a more sophisticated solution, e.g. have a `.env-ci` that would get put into place whine running in Go, but decided that the simpler the better here. If we need to customize the env vars for the Go runs we can do this in the actual job in Go.

tomas-stefano:
Yes, we can address this .env-ci in a different PR. I am just worry about
too much tech debts that we're facing lately. We need to address this tech
debts on the board/TP. But let's talk tomorrow about it.

2016-02-22 12:01 GMT+00:00 Misha Gorodnitzky <notifications@github.com>:

> @tomas-stefano <https://github.com/tomas-stefano> Yes, it turns out that
> the tests in Go have a problem if the .env is not there. :) We could come
> up with a more sophisticated solution, e.g. have a .env-ci that would get
> put into place whine running in Go, but decided that the simpler the better
> here. If we need to customize the env vars for the Go runs we can do this
> in the actual job in Go.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/moneyadviceservice/cms/pull/317#issuecomment-187145000>
> .
>



-- 
Atenciosamente,
Tomás D'Stefano

.env
misaka: @Tr4pSt3R What do you mean? The re-ordering? Yes, @habitullence has a keen sense of order and likes things to be well-ordered and easily searchable, quite understandably imo.

Tr4pSt3R: @misaka no I meant this line(#14) `MAS_CMS_URL="http://localhost:3000/"`

pmcostello:
:+1: looks good, let's get it merged and tested :)

.env
habitullence: Another vain attempt to organise the universe :laughing: What do you mean @Tr4pSt3R? My code is usually fairly intentional :stuck_out_tongue_winking_eye:

Tr4pSt3R: @habitullence, raising awareness for  line(#14) `MAS_CMS_URL="http://localhost:3000/"`. It is okay then let's merge and be `h=appy`

habitullence:
I agree that we might need to do something about the environment at some point but it's hard to know exactly what that is at this point as tests work fine with dummy data.

One option would be to tie this into the idea of having local Vagrant environments and encrypt keys for dependencies, like we do with existing environments.

Another option would be to mock external stuff.

My guess is, depending on what kind of testing we want, a combination of the two would be ideal.

misaka:
@tomas-stefano Just to comment on the technical debt question, I'm not sure this would be techn debt as we don't really know what it is we should be doing here at this time. We do know we shouldn't have production/sensitive data here, and that's already removed.

tomas-stefano:
Yes, I mean by tech debt, to have an env file with just the names and we need to filled with the data, but be careful to not commit this info. Yes, I agree with Rick said, we could see a way to use local vagrant for development then everything will be on your place.

PR #318, opened 2016-02-22
Add footer layout and page

mconnell:
__This PR compliments work done in__ https://github.com/moneyadviceservice/frontend/pull/1392

These changes introduce a new "Footer" layout to the CMS, along with a page using this layout. Footer pages have a different layout to only include the fields it uses, similar to the home page.

screen shot 2016-02-23 at 16 43 55


rnalexander:
Same question as on the other PR, how does it handle Welsh?

habitullence:
I'm looking for things to review but this looks like WIP, correct assumption?

paulanthonywilson:
@habitullence @rnalexander Why do we think this is WIP?

KateWilkinson:
Hi @rnalexander - my understanding is that a Welsh version of the footer page is created at the same time the English version is created (in the same way that a Welsh version of an article page is generated) so the Welsh text would just be input into/edited in that version.

paulanthonywilson:
@rnalexander What @KateWilkinson said. Looks like the English & Welsh pages are added around here: https://github.com/moneyadviceservice/cms/blob/add_footer_layout_and_page/lib/cms/page_builder.rb#L26-L27

paulanthonywilson:
:+1:

tomas-stefano:
If you dont mind I would like to pull locally and see this in more details.

app/views/pages/_form_for_footer.html.haml
```diff
@@ -0,0 +1,35 @@
...
+    - else
+      = render partial: 'form_blocks'
+
+    = render 'scheduler', object: form
+    = render 'footer_bar', object: form
+    = render 'comfy/admin/cms/partials/page_form_after', object: form
+
+= render 'activity_log_box'
+= render 'insert_link_dialog'
+= render 'insert_image_dialog'
```
tomas-stefano: This is equal to https://github.com/moneyadviceservice/cms/blob/master/app/views/pages/_form_for_home_page.html.haml

tomas-stefano: Can we use a partial? Or maybe render the same template and rename both.

app/views/pages/form_blocks/_footer.html.erb
```diff
@@ -0,0 +1,67 @@
+<%
+  namespace = { }
+  cms_object = @page || cms_object
+  tags = cms_object.tags(true).select{|t| t.is_cms_block?}.uniq{|t| t.identifier}
+
+  tags.each do |tag|
+    namespace[tag.namespace || 'default'] ||= []
+    namespace[tag.namespace || 'default'] << tag
+  end
+%>
```
tomas-stefano: Do we need this?

app/views/pages/form_blocks/_footer.html.erb
```diff
@@ -0,0 +1,67 @@
...
+            <div class="l-split-content">
+              <div class="l-split-content__col">
+                <h2>Contact Panel</h2>
+                <%= tag_for_identifier(tags, 'raw_contact_heading', cms_blocks, 5) %>
+                <%= tag_for_identifier(tags, 'raw_contact_introduction', cms_blocks, 6) %>
+                <%= tag_for_identifier(tags, 'raw_contact_phone_number', cms_blocks, 7) %>
+                <%= tag_for_identifier(tags, 'raw_contact_additional_one', cms_blocks, 8) %>
+                <%= tag_for_identifier(tags, 'raw_contact_additional_two', cms_blocks, 9) %>
+                <%= tag_for_identifier(tags, 'raw_contact_additional_three', cms_blocks, 10) %>
+                <%= tag_for_identifier(tags, 'raw_contact_small_print', cms_blocks, 11) %>
```
tomas-stefano: What it is tag_for_identifier?

app/views/pages/form_blocks/_footer.html.erb
```diff
@@ -0,0 +1,67 @@
...
+          <% namespace.each_with_index do |(name, tags), index| %>
+            <div<%= index == 0 ? ' class="active"' : '' %>>
+              <%= link_to name.humanize, "#ns-#{name}", :data => {:toggle => 'tab'} %>
+            </div
+          <% end %>
+        <% end %>
+
+        <div class="l-home-edit">
+          <% block_index = 0 %>
+          <% namespace.each_with_index do |(name, tags), index| %>
```
tomas-stefano: Why we need the name variable?

tomas-stefano: And why we need the index?

app/views/pages/form_blocks/_footer.html.erb
```diff
@@ -0,0 +1,67 @@
...
+                <%= tag_for_identifier(tags, 'raw_newsletter_heading', cms_blocks, 12) %>
+                <%= tag_for_identifier(tags, 'raw_newsletter_introduction', cms_blocks, 13) %>
+              </div>
+            </div>
+          <% end %>
+        </div>
+      <% end # fields_for %>
+    <% end %>
+  </div>
+</div>
```
tomas-stefano: Now I understand. This file looks like the app/views/pages/_form_blocks.html.haml that looks like  app/views/pages/forms/_home.html.erb. This view was a legacy template and it is copied in two places and now it will be a third. I would try to understand what this code are trying to solve, before copying to another place.

paulanthonywilson:
I'm *now* putting this back to WIP, while we look at @tomas-stefano's comments.

app/views/pages/form_blocks/_footer.html.erb
KateWilkinson: Hi @tomas-stefano, can I just check, when you say it's a legacy template, do you mean that there's a newer way of dealing with this?

tomas-stefano: No, unfortunately, there is no newer way. I just want to avoid add things that nobody understand what it does. And if we understand the root cause of this view, maybe we will not need to have this view that is trying to be too smart.

KateWilkinson:
@tomas-stefano I've tried to remove a lot of the duplication from the views - for example we no longer have the partials ```form_for_footer``` and ```form_for_home_page``` - instead we use the one general ```form``` and render the necessary content according to the layout type using a new helper method. 

I've also taken out the duplicate code from ```form_blocks/footer``` & ```form_blocks/home_page```, so it now only exists in ```form_blocks``` (which then renders those layout specific partials). 

Would you mind taking another look and letting me know what you think? Thanks for your help.

tomas-stefano:
Yes. I will take a look

app/helpers/pages_helper.rb
```diff
@@ -27,22 +27,15 @@ def page_slug(site_label, page_type_url, slug)
...
-    end
-  end
-
   def display_additional_button_menu?(page, user)
     return false if user.editor? || page_state_buttons(page).empty?
     page.publishable?
   end
 
+  def display_metadata_form_fields?(page)
+    page.layout.identifier != 'home_page' && page.layout.identifier != 'footer'
```
tomas-stefano: Can you change to an array? We are adding more layouts that have different forms and an array will be ideal. E.g: `!page.layout.identifier.in?(%w(home_page footer))`

KateWilkinson: @tomas-stefano done!

tomas-stefano:
:+1:

urfolomeus:
:+1:

KateWilkinson:
@tomas-stefano @urfolomeus @paulanthonywilson thanks everyone!

PR #322, opened 2016-02-29
Use env var mas_public_website_url

misaka:
Requires [PR 548](https://github.com/moneyadviceservice/scripts/pull/548) in scripts repo.

We don't want hard-coded values in configs if we can avoid them, this change allows the public website URL to be set via configuration in Puppet.


pmcostello:
:+1: looks good

jongilbraith:
@misaka now that the logic's gone looks like there's no need for the separate assignment and you can just use `ENV['MAS_PUBLIC_WEBSITE_URL']` directly on line 7.

misaka:
@jongilbraith Good point, missed that!

jongilbraith:
@misaka sweet, :+1: other than that :smile:

misaka:
@jongilbraith, @pmcostello Would you mind having another look? I tried cleaning up the env vars used here, `FRONTEND_URL` was 1. a miss-named (it's a domain, not a URL), and 2. not the name we would set in Puppet.

Would you mind sense-checking this change with the [associated one in the scripts repo](https://github.com/moneyadviceservice/scripts/pull/548)?

jongilbraith:
:+1: 

Nice to have been able to get rid of that fetch entirely.

PR #324, opened 2016-03-10
[DEV APPROVED]Extensive rework of the publishing workflow

jongilbraith:
This is a big rework of the publishing workflow for the CMS.

The existing system has all sorts of issues. The state system is confusing and inconsistent and that gets exposed in the interface which has all sorts of gotchas and quirks that make the users have to understand way too much about the internal functionality to do what they need.

The changes in this PR completely rework this so that from a user perspective there is a logical system for publishing, updating and scheduling pages.

Fundamentally the way all the data is structured is not easy to work with and I didn't want to tackle anything as big as reorganising that, so I've just tried to clean up what I can with that to make it easier to work with.

A summary of the main elements of the changes (some in the comfy gem in [this PR](https://github.com/moneyadviceservice/comfortable-mexican-sofa/pull/21)):

* Reorganised the state machine to reduce the number of needed events and made the purpose of the existing ones clearer.
* Established a set of pseudo states which are presented to users, wrapping the internal more specific states. E.g. 'published' which takes into account published posts, published posts with draft updates, scheduled posts that have gone live, etc.
* Added some classes to encapsulate the logic for retrieving, updating and removing content. This involves a concept of "Main content" and "Alternate content". For instance, when you have a published article and you want to make an update, you create a new draft which is referred to as the "alternate" content and exists until you publish it (and it replaces the existing live content) or if you schedule it to go live in the future, when it that happens.
* The addition of a foreign key to denote an "active revision". Previously, if there was a published article with an update being worked on, the live article was found by examining all the revisions and looking for the first article that fulfilled a set of criteria. This is much clearer and can be made use of in queries.
* A ton of comments to hopefully make things easier for people to understand in the future
* Much more extensive test coverage, including features. I'd like to add some request specs for the various publishing workflows but for now, this needs to be shipped.

That's about it.

Should also mention that I did go through an earlier WIP version of this with @tomas-stefano and generally got thumbs from him as to my general approach.


app/controllers/application_controller.rb
```diff
@@ -8,4 +8,5 @@ def check_admin
   def english_site
     @english_site ||= Comfy::Cms::Site.find_by(label: 'en')
   end
+  helper_method :english_site
```
habitullence: Would be nice to have a line break here.

app/helpers/page_blocks_helper.rb
```diff
@@ -0,0 +1,7 @@
+module PageBlocksHelper
+
```
habitullence: Excess line breaks.

app/models/alternate_page_blocks_register.rb
```diff
@@ -0,0 +1,76 @@
+# This class wraps logic for updating the "alternate" blocks of a page, used
+# when there is a published article with an "alternate" version, i.e. a new
+# version in draft or that has been scheduled to go live in the future.
+#
+# See the comments atop of PageBlocksRegister for a deeper explanation.
+#
+class AlternatePageBlocksRegister
+
```
habitullence: Don't think we usually have line break at start of class.

app/models/alternate_page_blocks_register.rb
```diff
@@ -0,0 +1,76 @@
...
+    if @new_blocks_attributes.is_a?(Hash)
+      @new_blocks_attributes.values
+    else
+      @new_blocks_attributes
+    end
+  end
+
+  def save!
+    if page.published_being_edited?
+
```
habitullence: Excess LB.

app/controllers/pages_controller.rb
```diff
@@ -1,8 +1,65 @@
...
+  before_action :check_alternate_available, only: [:edit, :update, :destroy]
+  before_action :check_can_destroy, only: :destroy
+
+  def index
+    @all_pages = Comfy::Cms::Page.includes(:layout, :site, :categories)
+    @pages_by_parent = pages_grouped_by_parent
+    @pages = apply_filters
+  end
+
+  def new
```
habitullence: Is this needed?

habitullence:
Big PR, done my best to understand what's going on. Looks good, good test coverage :+1:

rnalexander:
Agreed, very happy to see so much test coverage.  Do we know what's going on with the CodeClimate failure?
Otherwise :+1:

app/controllers/pages_controller.rb
jongilbraith: As discussed offline, I include an empty method to make it easy to see that this action does get used without having to refer to routes/views, plus also with this controller inheriting from another controller you don't need to go look in the parent to check the exact behaviour.

jongilbraith:
As discussed on Slack, the main thing in Code Climate is highlighting some of the long/complex methods. These unfortunately IMHO are unavoidable for us because of the complexity of the solution required to deal with the weird way we store content in the database.  Had the schema been created with revisioning and scheduling in mind right from the start them none of this would be needed, but it doesn't look like that was the case. :smile:

jongilbraith:
Cheers @habitullence @rnalexander.

Branch in comfy has been merged and have updated the commit here to pull in the new gem.

All ready to go to merge on Monday and go to QA.

PR #327, opened 2016-03-24
7222 fixes to scheduling functionality

jongilbraith:
It was discovered that the new publishing workflow fell down when trying to be used on non article type pages (home page, footer).  This PR addresses the issues and now they can be updated and scheduled in the same manner as articles.

The solution is a bit hacky, but that fits in perfectly with the way these forms are built.


rnalexander:
:+1:

habitullence:
:+1:

PR #329, opened 2016-05-10
[DEV APPROVED]7228 link to mas bug

misaka:
The tests added here include a certain amount of refactoring to partially take out the reliance on a singleton object `cms_page`. This allows us to work with two pages in the same test, for example when linking to pages like in this change. The end goal should probably be to move the fixture generation into factory girl ... although that's probably for later change.

### DIFF
--- a/app/controllers/links/pages_controller.rb
+++ b/app/controllers/links/pages_controller.rb
@@ -1,6 +1,6 @@
 class Links::PagesController < PagesController
   def index
-    pages  = Comfy::Cms::Page.includes(:site).published.filter(params.slice(:category))
+    pages  = Comfy::Cms::Page.includes(:site).where('state != ?', 'unpublished').filter(params.slice(:category))
     pages  = Comfy::Cms::Search.new(pages, params[:search]).results if params[:search].present?
 
     @pages = PageMirror.collect(pages)
--- a/app/controllers/links/pages_controller.rb
+++ b/app/controllers/links/pages_controller.rb
@@ -1,6 +1,6 @@
 class Links::PagesController < PagesController
   def index
-    pages  = Comfy::Cms::Page.includes(:site).where('state != ?', 'unpublished').filter(params.slice(:category))
+    pages  = Comfy::Cms::Page.includes(:site).where.not(state: :published).filter(params.slice(:category))
     pages  = Comfy::Cms::Search.new(pages, params[:search]).results if params[:search].present?
 
     @pages = PageMirror.collect(pages)
--- a/features/step_definitions/edit_page_steps.rb
+++ b/features/step_definitions/edit_page_steps.rb
@@ -18,11 +18,27 @@ Then(/^I should not be able to see live draft article$/) do
 end
 
 Given(/^I have an articles with unpublished changes$/) do
-  edit_page.load(site: cms_site.id, page: cms_new_draft_page.id)
+  load_page_in_editor(site: cms_site, page: cms_new_draft_page).slug
   edit_page.content.set("New Published Content")
   edit_page.publish.click
   edit_page.content.set("New unpublished Content")
   edit_page.create_new_draft.click
+end
+
+Given(/^I have a scheduled article with the slug "([\w\s-]+)"/) do |title|
+  build_cms_scheduled_page(identifier: title)
+end
+
+Given(/^I have a published article "([\w\s-]+)" in "([a-zA-Z]{2})"$/) do |slug, locale|
+  build_cms_published_page(identifier: slug, locale: locale)
+end
+
+Given(/^I have a scheduled update to the "([a-zA-Z]{2})" mirror of "([\w\s-]+)"$/) do |slug, locale|
+  build_cms_scheduled_page(identifier: slug, locale: locale)
+end
+
+Given(/^the published article has a "([a-zA-Z]{2})" update scheduled$/) do |locale|
+  cms_scheduled_page(identifier: title)
 end
 
 When(/^I view the live published article$/) do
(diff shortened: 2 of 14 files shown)


rnalexander:
:+1:

tomas-stefano:
👍 (After fixing the code climate warning)

features/step_definitions/edit_page_steps.rb
```diff
@@ -142,3 +163,18 @@
...
+Then(/^I should see "(.*?)" in the list of linkable pages/) do |name|
+  @linkable_page = edit_page.link_manager.pages.results.find do |row|
+    row.page_name.text.include? name
+  end
+  expect(@linkable_page).to_not be_nil
+end
+
+Then(/^the "([a-zA-Z]{2})" and "([a-zA-Z]{2})" versions are available for linking$/) do |*locales|
+  locales.each do |locale|
+    expect(@linkable_page.send(locale)).
```
jiannis: Using `@linkable_page` as an instance variable initialised in the previous step makes these two step definitions dependent. Maybe a helper method instead of a `@linkable_page` instance variable?

features/support/env.rb
```diff
@@ -8,7 +8,7 @@
 Cucumber::Rails::Database.javascript_strategy = :truncation
 
 Capybara.register_driver :poltergeist do |app|
-  Capybara::Poltergeist::Driver.new(app, js_errors: true, inspector: true)
+  Capybara::Poltergeist::Driver.new(app, js_errors: true, inspector: true, window_size: [2048, 1536])
```
jiannis: Why do need this specific window size?

features/support/hooks.rb
```diff
@@ -0,0 +1,17 @@
...
+    unless saved_image.empty?
+      puts "saved screenshot for scenario \"#{scenario.name}\" to #{saved_image}"
+    end
+    saved_page = save_page("#{filename}.html")
+    unless saved_page.empty?
+      puts "saved page for scenario \"#{scenario.name}\" to #{saved_page}"
+    end
+  end
+end
+
```
jiannis: 🎩

features/support/world/cms.rb
```diff
@@ -13,11 +13,15 @@ def cms_sites
...
-      self.layout ||= cms_site(locale).layouts.create!(identifier: identifier, content: '{{ cms:page:content:rich_text }}')
+      self.layout ||= cms_site(locale).layouts.create!(
+        identifier: 'article',
+        content: '{{ cms:page:content:rich_text }}'
+      )
     end
 
     def cms_root(locale='en')
       self.root ||= cms_site(locale).pages.create!(layout: cms_layout, label: 'root')
+      self.root
```
jiannis: Doesn't the previous statement return `self.root` ?

app/controllers/links/pages_controller.rb
```diff
@@ -1,6 +1,6 @@
 class Links::PagesController < PagesController
   def index
-    pages  = Comfy::Cms::Page.includes(:site).published.filter(params.slice(:category))
+    pages  = Comfy::Cms::Page.includes(:site).where('state != ?', 'unpublished').filter(params.slice(:category))
```
jiannis: Could we do the same thing with a `scope` in [`Comfy::Cms::Page`](https://github.com/moneyadviceservice/cms/blob/master/app/models/comfy/cms/page.rb) ?

```ruby
class Comfy::Cms::Page < ActiveRecord::Base
  scope :allthethings, -> { where.not(state: 'unpublished') }
end

def index
  pages = Comfy::Cms::Page.includes(:site).allthethings.filter(params.slice(:category)
end
```

or maybe use the existing `unpublished` scope and try something like:

```ruby
def index
  pages = Comfy::Cms::Page.includes(:site).where.not.published.filter(params.slice(:category)
end
```

features/support/env.rb
misaka: Only to be able to see more of the page when saving a screenshot for debugging.

features/step_definitions/edit_page_steps.rb
misaka: This is a tricky one. Not having steps dependant is good design, but in this case it gets awkward if we want to bundle both tests together into `I should see "(.*?)" in the list of linkable pages` and `the "([a-zA-Z]{2})" and "([a-zA-Z]{2})" versions are available for linking` steps. The step wording gets awkward, etc. And it's nice to test this criteria separately instead of squishing them into the same step.

jiannis: 👌

features/support/env.rb
jiannis: 👌

app/controllers/links/pages_controller.rb
misaka: Looks good! Let me test this.

misaka:
@tomas-stefano re: the [Code Climate issue in `UI::Pages::Edit`](https://codeclimate.com/repos/55798a3869568024020134d1/UI::Pages::Edit?from_sha=cad65949&to_sha=26d5331d), after reviewing the change we discussed, I think it's better to leave this as-is. As we saw, sections are stored in a separate directory in our structure, and if they're not being re-used I don't really see enough benefit to warrant moving them so far away from the code where they're being used. Do you agree?

tomas-stefano:
I agree in terms. The structure it is not flexible and is totally awful. Sections belongs to pages, so it should not be separated. And there is the tag section duplication either on tag.rb and header.rb. Confusing. **But** we should create a card to improve this structure (or add this info on the factory girl replacement on the cucumber spec). If this is added on the user story, I am totally fine. Otherwise, I am not.

misaka:
@tomas-stefano I've cleaned up the code duplication that we saw in Code Climate, and added a couple TP tickets: [7351](https://moneyadviceservice.tpondemand.com/entity/7351) and [7353](https://moneyadviceservice.tpondemand.com/entity/7353). Of course, Code Climate is still bugging about some other duplication but I reckon that that can wait until the fixture refactoring work.

tomas-stefano:
👍 👍

rnalexander:
👍  :shipit:

jongilbraith:
@misaka nice job with the tests, I was frustrated with the state of them last time I was working on this but didn't have it in me to do much to improve them. Still plenty of scope for improvement but your changes address the main sticking points nicely.

👍 from me.

UmairIft:
All looks good 👍
had a mini discussion with @jongilbraith , the issue with scheduled publishing had to do with UTC time zones. The scheduled time is -1hr of current time . A fix for another ticket

PR #334, opened 2016-08-11
[DEV APPROVED] Rename azure asset container env variable

misaka:
This brings it in line with the name of other azure asset container
variables as defined in scripts.

### DIFF
--- a/.env
+++ b/.env
@@ -2,7 +2,7 @@
 # N.B. This file is intended as a placeholder. Never put any real values in this file and commit.
 ####################################################################################################
 
-AZURE_ASSETS_STORAGE_CMS_ACCOUNT_CONTAINER='container'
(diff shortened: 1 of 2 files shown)


misaka:
This needs to be deployed at the same time as the changes in this PR: https://github.com/moneyadviceservice/scripts/pull/691

tomas-stefano:
👍

jongilbraith:
👍

tomas-stefano:
I have a good and bad news!  

Bad news: The .env file is checked and is used on the tests. 

Good news: So I added a commit renaming the env variable there too here https://github.com/moneyadviceservice/cms/pull/334/commits/cfce7eba0e1c6286e81449a0af9f420ef7f0b759 and this will fix this: https://go.dev.mas.local/go/pipelines/cms_commit_preview/16/test_and_build/1

PR #335, opened 2016-10-06
When uses CMS locally, CMS SHOULD NOT upload to Azure

tomas-stefano:
The CMS locally was using the Azure config for uploading.

So the title was pretty obvious on the expect behaviour.


slaywell:
👍

Tr4pSt3R:
👍

PR #337, opened 2016-10-07
Removing bad Karma tests

slaywell:
The Karma tests for DataCharts rely on checking properties of a JS object that is returned from the Google API - this object changes every few months. We are not testing a documented part of their API with these tests (nor are we actually checking that the DataChart is successfully created).

To be exact, we _could_ fix this right now by making this change:
from:
`expect(dataChart.Ka.chart.container).equal('some-chart-container');`
to:
`expect(dataChart.g$.chart.container).equal('some-chart-container');`

But that is merely pushing the problem on a few months until this happens again (see the commit history of this file). `g$` is not documented anywhere.

At the moment we can't get the CMS pipeline to run. So I'm suggesting we remove these tests, and write alternative ones that actually check the chart has rendered.

### DIFF
--- a/spec/javascripts/tests/EmbeddedGoogleAnalytics_spec.js
+++ b/spec/javascripts/tests/EmbeddedGoogleAnalytics_spec.js
@@ -38,7 +38,10 @@ describe('EmbeddedGoogleAnalytics', function() {
       expect(component.config.path).to.equal('/en/articles/foo');
     });
 
-    it('create a DataChart', function() {
+    // Skipping next three tests. 
+    // They are checking a JS object that is returned from Google API, which changes every few months without warning / documentaiton
+    // Instead we should be testing that the chart actually renders
+    xit('create a DataChart', function() {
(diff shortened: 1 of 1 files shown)


pmcostello:
best to remove these for the time being, until we can get them working locally 👍

jongilbraith:
Agree with @pmcostello, just remove them entirely rather than comment out.

👍  otherwise

Tr4pSt3R:
Spot on. Writing a test to check if the chart has been successfully rendered should be sufficient. 
👍 👍

slaywell:
@jongilbraith we can skip the tests rather than deleting them (see latest commit)? I prefer to leave a visible reminder that we should add them again.

jongilbraith:
@slaywell cool, that works for me. Maybe a brief comment about why they've been disabled might be worthwhile too.

I just don't like committing commented out code on the basis of "we still need this and plan to come back to it" as you should be leaning on git to retrieve such things.

PR #338, opened 2016-10-07
[DEV APPROVED] - Page feedbacks API

tomas-stefano:


jriga:
LGTM 👍

db/schema.rb
```diff
@@ -199,6 +199,20 @@
     t.string  "locale",        null: false
   end
 
+  create_table "page_feedbacks", force: true do |t|
+    t.integer  "page_id"
+    t.boolean  "liked",      default: true
```
Tr4pSt3R: The page is "Liked" by default?

Tr4pSt3R:
👍

db/schema.rb
tomas-stefano: I prefer to choose one by default, so either is false or true. It doesn't matter anyway because it is forced by the API 'true' or 'false'.

jongilbraith:
LGTM 👍

PR #339, opened 2016-10-10
7517 - remove old migrations

jongilbraith:
This PR addresses [this ticket](https://moneyadviceservice.tpondemand.com/entity/7517) which raised issues running the migrations on a new db.

In actual fact the migrations just shouldn't have been attempted at all, with the schema instead being loaded direct from `schema.rb`.  The README states to run `bin/setup`, which does do this.

So ultimately the issue raised in the ticket isn't a bug. This is however a good prompt to do some cleanup and clear out the old migrations.


tomas-stefano:
👍

pmcostello:
👍

PR #340, opened 2016-10-11
Adding the mas-version-diff script to CMS

slaywell:
In the Frontend repo we have the `bin/mas-version-diff.sh`script which we use to compare Frontend versions before deploying. The output allows us to see all commits that are to be included in the release.

This doesn't currently exist in the CMS project so with this PR I suggest we add it?

### DIFF
--- a/bin/mas-version-diff
+++ b/bin/mas-version-diff
@@ -1,0 +1,14 @@
+#!/usr/bin/env sh
+
+# USAGE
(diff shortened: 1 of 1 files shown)


johnplayer1982:
I dont think we need this as we can do it from within scripts

eg:
```
./mas-version-diff 492az 493az cms
```

slaywell:
Cool, I wondered how others were doing this.

PR #342, opened 2016-10-12
[DEV APPROVED] 7674 - Extract a list of All articles and what categories they sit in

jriga:
**Feature**
Added task `rake adhoc:extract_page` to create csv file for published articles and all their categories

### DIFF
--- a/lib/tasks/adhoc.rake
+++ b/lib/tasks/adhoc.rake
@@ -3,7 +3,8 @@ namespace :adhoc do
   task :extract_page => :environment do
     File.open('pages_with_cat.csv','w') do |file|
       file.write "Url\tCategories\n"
-      Comfy::Cms::Page.where(state: :published).find_each do |page|
+      en = Comfy::Cms::Site.where(label: :en).first
+      Comfy::Cms::Page.where(site: en, state: :published).find_each do |page|
         file.write "/en/articles/#{page.slug}\t#{page.categories.map(&:label).join(', ')}\n"
         $stdout.write '.'
       end
--- a/lib/tasks/adhoc.rake
+++ b/lib/tasks/adhoc.rake
@@ -1,0 +1,14 @@
+namespace :adhoc do
+  desc 'Extract a list of All articles and what categories they sit in'
+  task :extract_page => :environment do
+    File.open('pages_with_cat.csv','w') do |file|
+      file.write "Url\tCategories\n"
+      Comfy::Cms::Page.where(state: :published).find_each do |page|
+        file.write "/en/articles/#{page.slug}\t#{page.categories.map(&:label).join(', ')}\n"
+        $stdout.write '.'
+      end
+      puts "\nCreated csv file `#{file.path}`"
+      file.flush
+    end
+  end
+end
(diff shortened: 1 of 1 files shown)


lib/tasks/adhoc.rake
```diff
@@ -0,0 +1,14 @@
+namespace :adhoc do
+  desc 'Extract a list of All articles and what categories they sit in'
+  task :extract_page => :environment do
+    File.open('pages_with_cat.csv','w') do |file|
+      file.write "Url\tCategories\n"
+      Comfy::Cms::Page.where(state: :published).find_each do |page|
+        file.write "/en/articles/#{page.slug}\t#{page.categories.map(&:label).join(', ')}\n"
```
jongilbraith: I think this is going to return both english and welsh pages, so you probably want some logic to populate the right locale.

tomas-stefano: Agree with @jongilbraith. Probably you want: 

```ruby
file.write "/#{page.site.locale}/articles/#{page.slug}\t#{page.categories.map(&:label).join(', ')}\n"
```

Explaining: You can do `english  = Comfy::Cms::Site.find_by(label: 'en')`or `welsh = Comfy::Cms::Site.find_by(label: 'cy')`
And then you can access using `english.pages` or `welsh.pages`.
The site object has *locale* or *label* as attribute to get the right locale. You choose.

jriga: ♻️

lib/tasks/adhoc.rake
```diff
@@ -0,0 +1,15 @@
+namespace :adhoc do
+  desc 'Extract a list of All articles and what categories they sit in'
+  task :extract_page => :environment do
+    File.open('pages_with_cat.csv','w') do |file|
+      file.write "Url\tCategories\n"
+      en = Comfy::Cms::Site.where(label: :en).first
```
jongilbraith: `Comfy::Cms::Site.find_by(label: :en)`?

lib/tasks/adhoc.rake
```diff
@@ -0,0 +1,15 @@
...
+      Comfy::Cms::Page.where(site: en, state: :published).find_each do |page|
```
jongilbraith: Might as well use associations here, `en.pages.where(state:published)`

jriga: I use `find_each` because it uses `find_in_batches` underneath which is suited for this type of processing.

jongilbraith: But.... I wasn't commenting on the use of `find_each`? My comment was about how you derived the scope before you called `find_each` on it.

So in full it would be: `en.pages.where(state:published).find_each do |page|`

And now I come to think if it, you need to use the `published` scope here, not merely selecting based on the state column.

So: `en.pages.published.find_each do |page|`

There's SOOOO much more to how you find published articles! (https://github.com/moneyadviceservice/cms/pull/324)  #stilltraumatised #neverforget

lib/tasks/adhoc.rake
jongilbraith: I know they're the same and I'm not precious about it either, but if you have a tool made specifically for a particular job seems to me you might as well use it when the situation arises.

lib/tasks/adhoc.rake
jriga: :no_mouth:
thanks @jongilbraith for the context

jongilbraith:
👍 👍

tomas-stefano:
👍

PR #344, opened 2016-10-20
7695 - Fix bug when inserting external link with no anchor text

jongilbraith:
When inserting an external link via the editor, if you do this without selecting some anchor text, the inserted kramdown markup appears over two lines/paragraphs.

This fixes that bug.


tomas-stefano:
👍  👍

rnalexander:
:+1: :shipit:

PR #345, opened 2016-10-21
7663 return page feedback count

jriga:
**Feature**
Returns page feedback counts for like and dislike after creating feedback.
[7663](https://moneyadviceservice.tpondemand.com/entity/7663)

### DIFF
--- a/app/controllers/api/page_feedbacks_controller.rb
+++ b/app/controllers/api/page_feedbacks_controller.rb
@@ -1,4 +1,6 @@
 class API::PageFeedbacksController < APIController
+  before_action :find_page
+
   def create
     page_feedback = PageFeedback.new(page_feedback_params)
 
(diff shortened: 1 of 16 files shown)


jriga:
👁 @tomas-stefano

tomas-stefano:
👍 👍 Sounds great but IMHO will be better to merge on the branch that I worked. Wdyt?

app/controllers/api/page_feedbacks_controller.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  def find_page
+    @page = @current_site.pages.find_by(slug: params[:slug])
+
+    render json: { message: %(Page "#{params[:slug]}" not found) },
+           status: 404 if @page.blank?
+  end
+
+  def page_feedback_params
+    params.permit(:liked, :session_id).merge(page_id: @page.id)
```
jongilbraith: This is a bit unusual... Rather than adding `page_id` to the params, I'd expect whatever depends on it to be referring to `@page`.

tomas-stefano: Yes, it is true! I will address on the proper PR here: #338

jriga:
oops I meant to create the PR against the branch `feature/feedback-like-dislike-articles` so that it could just be merged into it.
Creating #346 to fix that

PR #349, opened 2016-10-25
Fix issue of assets not being show in development

tomas-stefano:
* When you use a different port of CMS, the assets are not showed
  anywhere. Removing this assets configuration solves the issue.


jriga:
👍

jongilbraith:
👍

PR #351, opened 2016-11-04
7686 - A11Y - a way to add captions to tables

jongilbraith:
Explored various different approaches and eventually went with leaving tables to be generated as normal, and adding a post processor to look for a paragraph with class of `caption` immediately after a table and if found add it as the `<caption>` tag.

Adding a class to a paragraph is done like so:

```
A paragraph of text
{: .a-class-to-apply}
```

The javascript table insert snippet has been updated to include the caption paragraph.

I ran the styling past @slaywell and we ended up with leave it as it was as it appears that the styling had already been considered in the past.


rnalexander:
:+1:

tomas-stefano:
👍

PR #355, opened 2016-11-14
[DEV APPROVED] - 7711 protect cms endpoints

tomas-stefano:
Background:
 
On the development of this feature it was added one endpoint with two HTTP verbs on CMS API:
 
POST   "/api/en/example-article/page_feedbacks"
PATCH "/api/en/example-article/page_feedbacks"
 
The problem that the CMS API is public and doesn't have any authentication mechanisms, so anybody malicious can create a script to just create a lot of likes and dislikes for each article if they want it. So adding an authentication solves the problem.

### DIFF
--- a/app/filters/api/authentication_filter.rb
+++ b/app/filters/api/authentication_filter.rb
@@ -1,0 +1,16 @@
+module API
+  class AuthenticationFilter
+    MAS_CMS_API_TOKEN = ENV.fetch('MAS_CMS_API_TOKEN')
+
+    def self.before(controller)
+      authenticate?(controller)
+    end
+
+    def self.authenticate?(controller)
+      controller.authenticate_or_request_with_http_token do |token, _|
+        ::Digest::SHA256.hexdigest(MAS_CMS_API_TOKEN) ==
+          ::Digest::SHA256.hexdigest(token)
+      end
+    end
+  end
+end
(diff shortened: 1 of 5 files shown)


Tr4pSt3R:
👍 looks good. I can't spot any issue

tomas-stefano:
@Tr4pSt3R sorry to be patronising ... but try "not just spot issues". Instead tries to reviews see if you're satisfied to this code to be on production and if you're agree that this code to be maintained by us AND see if makes sense to your logical perspective. IMHO, spot issues is just a consequence of this.

app/filters/api/authentication_filter.rb
```diff
@@ -0,0 +1,16 @@
+module API
+  class AuthenticationFilter
+    MAS_CMS_API_TOKEN = ENV.fetch('MAS_CMS_API_TOKEN')
```
jriga: What happens if MAS_CMS_API_TOKEN is `nil`. Should it raise an error or silently rejected request

tomas-stefano: Good point. I would say the same as frontend behaviour. The same comment: https://github.com/moneyadviceservice/frontend/pull/1599#discussion_r88065107

jriga:
LGTM 👍

PR #357, opened 2016-11-22
[WIP] 7749/7755 - SPIKE - navigation categories active yaml

jongilbraith:
Opening a PR for this to start some discussion.

My understanding of the requirements is that the "Clumps" that are the main groupings for the nav simply contain categories, with those subcategories being used to generate the lists of links under each heading on the different drop down pages.

I've populated them with categories as close to the ones described in the ticket, but it's assumed that when we progress with this, part of the work will include some reorganisation of the existing categories.

I've used the [ActiveHash](https://github.com/zilkey/active_hash) library as the basis for the clumps here, but also explored an ActiveRecord based approach [in another branch](https://github.com/moneyadviceservice/cms/tree/7755-navigation-categories-activerecord).

### DIFF
--- a/app/models/clump.rb
+++ b/app/models/clump.rb
@@ -1,0 +1,67 @@
+class Clump < ActiveHash::Base
+  self.data = [
+    {
+      id: 1,
+      name_en: 'Debt & Borrowing',
+      name_cy: 'Dyled a benthyca',
+      category_labels: [
+        'debt-and-borrowing'
+      ]
+    },
+    {
+      id: 2,
+      name_en: 'Homes & Mortgages',
+      name_cy: 'Cymorth gyda morgeisi',
+      category_labels: [
+        'homes-and-mortgages'
+      ]
+    },
+    {
+      id: 3,
+      name_en: 'Managing Money',
+      name_cy: 'Rheoli arian',
+      category_labels: [
+        'budgeting-and-managing-money',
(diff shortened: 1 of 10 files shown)


jongilbraith:
Included the json end point in this spike, so just realised I've covered [ticket 7749](https://moneyadviceservice.tpondemand.com/entity/7749) too.

app/models/clump.rb
```diff
@@ -0,0 +1,71 @@
...
+      name_en: 'Insurance',
+      name_cy: 'Yswiriant',
+      category_labels: [
+        'insurance'
+      ]
+    }
+  ]
+
+  def categories
+    category_labels.collect { |l| Comfy::Cms::Category.find_by(label: l) }
```
jriga: We could load everything in one go with `Comfy::Cms::Category.where(label: category_labels)`

jongilbraith: I initially went that route, but if the label doesn't exist (typo, moved, etc) then we'd just end up with it being skipped.

This way, we'll get a clear record not found thrown, which feels like a good safety check if we're administering things with a config file (i.e. that doesn't come with validation).

jriga: Make sense. 👍

jongilbraith:
Closing this in favour of the activerecord approach, and intention to smarten this up for a "real" PR.

PR #362, opened 2016-11-28
[DEV APPROVED] 7805 - "clumps" - models/schema changes, initial data, api, admin view

jongilbraith:
This is a formalising of the work spiked in ticket 7755.

* `Clump` model, consisting of names and descriptions in English and Welsh, plus an ordinal column
* `Clumping` model to associate clumps with categories, again with an ordinal column for ordering
* A migration to populate some initial clumps and categories to go in them. This populates the clumps as best as we can with the categories that currently exist and @jriga's next tranche of work will include a script to modify the categories and that will also take care of populating the clumps with the correct new categories
* An api end point at `/:locale/clumps.json`
* A page in the back end that displays the clump/category hierarchy

### DIFF
--- a/app/controllers/api/clumps_controller.rb
+++ b/app/controllers/api/clumps_controller.rb
@@ -1,0 +1,33 @@
+module API
+  class InvalidLocale < Exception;  end
+
+  class ClumpsController < ApplicationController
+    skip_before_action :load_cms_page
+    around_filter :validate_locale
+
+    def index
+      @clumps = Clump.all
+      render json: @clumps, scope: locale
+    end
+
+    private
+
+    def locale
+      fail InvalidLocale.new unless params[:locale].in?(accepted_locales)
+
+      params[:locale]
+    end
+
+    def validate_locale
+      yield
+    rescue InvalidLocale
+      render json: {
+        error: "Unaccepted locale (must be #{accepted_locales.join(' ')})"
+      }, status: 400
+    end
+
+    def accepted_locales
+      @accepted_locales ||= Comfy::Cms::Site.pluck(:path)
+    end
+  end
+end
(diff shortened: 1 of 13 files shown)


app/controllers/api/clumps_controller.rb
```diff
@@ -0,0 +1,33 @@
+module API
+  class InvalidLocale < Exception;  end
+
+  class ClumpsController < ApplicationController
```
tomas-stefano: Wdyt about inherit from APIController?

jongilbraith: Tbh it wasn't a conscious decision not to inherit from `APIController` - I used `CategoryContentsController` as a basis for this, and that inherits from `ApplicationController`.

Looking into it a bit further, looks like the reason for this is because `APIController` contains some functionality for looking up the site, whereas categories (and now clumps) aren't site specific, so don't make use of this behaviour.

That said though, this new controller does now introduce some duplication and I'm starting to wonder now if there's a better way to organise things (thinking maybe using concerns), so I'll have a look and see if I can come up with something.

tomas-stefano: Ok :+1: I am just asking because the admin is using ApplicationController. Any change on app controller will affect this api clumps controller.

tomas-stefano:
After the discussion, I conclude LGTM 👍  🍫

jongilbraith:
Sweet, thanks @tomas-stefano.

jriga:
👍

jongilbraith:
Closing this in favour of https://github.com/moneyadviceservice/cms/pull/364 which has been branched off the new global nav feature branch `global-nav-master`.

PR #365, opened 2016-11-30
[DEV APPROVED] - Delete all feedback when page is destroyed

tomas-stefano:
Otherwise they will leave orphan records.

### DIFF
--- a/app/models/comfy/cms/page.rb
+++ b/app/models/comfy/cms/page.rb
@@ -4,7 +4,10 @@ require Rails.root.join('lib', 'comfortable_mexican_sofa', 'extensions', 'is_tag
 class Comfy::Cms::Page < ActiveRecord::Base
   include ComfortableMexicanSofa::IsTaggable::ModelMethods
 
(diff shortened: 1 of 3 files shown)


jriga:
Looks good 👍 
However, I've got a question which is I think outside of the scope of this PR. 
Do you guys thinks a soft delete would be better than an hard delete? Soft delete enables us to look back at user feedbacks in future

tomas-stefano:
Yes, I talk with Umair and he said there is ok to delete all feedbacks when the page is destroyed. He will increase the awareness of the editors about it.

jongilbraith:
@jriga indeed, it would be interesting to find out when deletion is actually used.

They have the ability to 'unpublish' and I was under the impression that was the route that was followed when an article was removed late in it's life (if it's no longer relevant, for example) and deletion was mainly used for pre-publishing articles that they've decided not to go ahead with any more.

jongilbraith:
To clarify, my comment is just a general "useful/interesting to know" thing. If there's no desire to keep hold of the feedback for a deleted article then 👍 .

PR #367, opened 2016-12-01
[DEV APPROVED]Use azure 0.7.7 when using ruby 2.3.3

tomas-stefano:
* This fix the "Can not modify frozen String" when upload files
on the CMS.


jongilbraith:
👍

rnalexander:
:+1: :shipit:

PR #368, opened 2016-12-06
7758 - initial clump administration

jongilbraith:
This ticket is being split off into two tickets. It originally was for adding an interface for editing aspects of the clumps including the tool links, however we still need to do some backend work for those tool links, that portion is being put into a new ticket.  This PR now solely deals with adding some interface for the parts of the clumps which already exists (titles, descriptions, categories).

The interface I've implemented here largely mirrors that of the categories.

We can create, edit and reorder clumps (but, like with categories, not delete them).

We can reorder the categories in the clumps via the category#show page.

We can (un)assign categories to clumps via the category#edit page.

### DIFF
--- a/app/controllers/clumps_controller.rb
+++ b/app/controllers/clumps_controller.rb
@@ -1,7 +1,53 @@
 class ClumpsController < Comfy::Admin::Cms::BaseController
 
   def index
-    @clumps = Clump.all
+    @clumps = Clump.order('ordinal ASC')
+  end
+
+  def show
+    @clump = Clump.find(params[:id])
+  end
+
+  def new
+    @clump = Clump.new
+  end
+
+  def create
+    @clump = Clump.new(clump_params)
+    if @clump.save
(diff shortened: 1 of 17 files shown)


rnalexander:
:+1:

pmcostello:
👍 nice!

jongilbraith:
@rnalexander @pmcostello thanks both!

app/controllers/clumps_controller.rb
```diff
@@ -1,7 +1,53 @@
...
+      end
+      redirect_to clump_path(@clump)
+    rescue ActiveRecord::RecordInvalid
+      render :show
+    end
+  end
+
+  def reorder
+    Clump.transaction do
+      params[:order].split(',').each_with_index do |clump_id, index|
```
tomas-stefano: Can we move params[:order].split(',') to a private method and use here and the method above?

tomas-stefano:
👍 👍

PR #369, opened 2016-12-09
7831 - (tool) links in clumps

jongilbraith:
Had a few false starts on this, trying a few different approaches to try and avoid having such a similar model to the existing `Link` model but ended up just going with just that.

The spec for this requires that we can have up to four links in a clump, so in order to simplify the interface we just build four of these when you create a clump and then the interface allows you to edit these but not add or remove any.  Validation allows you to either leave some blank or requires you to populate them entirely, and when rendering the json we  only show the complete ones.

I tried various approaches to reuse the existing model but it wasn't suitable for STI (the subclass of Link would have to remove the validation applied in the parent) and when I tried adding conditionals based on the `linkable` to validation things started to get unnecessarily complex.

The starter clump information was pulled out of the static mock up @davidtrussler did, but I expect will be subject to change in the future.

Also included in the `ClumpLink` model is a `style` field which is intended to be used to style the links, as these links might not always be tools. This is currently free text, but I would hope we can define some permitted values and change the interface to a drop down.

### DIFF
--- a/app/controllers/clumps_controller.rb
+++ b/app/controllers/clumps_controller.rb
@@ -9,7 +9,9 @@ class ClumpsController < Comfy::Admin::Cms::BaseController
   end
 
   def new
-    @clump = Clump.new
+    @clump = Clump.new do |clump|
+      4.times { clump.clump_links.build }
+    end
   end
 
   def create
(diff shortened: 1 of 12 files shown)


tomas-stefano:
Great work! :+1:

jriga:
👍

jongilbraith:
@tomas-stefano @jriga thanks both

PR #372, opened 2016-12-13
[DEV APPROVED] Speed up CMS article requests

tomas-stefano:
# Description

Webfeeds request to the CMS on `/en/articles/published` and it takes almost 2 minutes to finish the whole request. You can see on TP as 7833.

Further investigation were found that there was a lot of N+1 queries and a lot of markdown processing to html. Detailed explanation is below.

So removing the n+1 queries and add a 'sort of cache' (convert the html only when you save the page and not when you request the page) will solve half of the problem.

These are the tasks to complete related to this PR:

- [x] Remove N+1 queries
- [x] Create the Comfortable Mexican sofa PR
- [x] Create the Mastalk PR
- [x] Cache the html of all articles (or in the database or in the cache)
- [x] Create a data migration to convert all markdown to html

## Consequences

*This will affect all apps that are using the CMS (the site and the webfeeds and probably universal credit).*

## Related PRs

This PR is waiting for two PR's to be merged. After the merge I will address on the Gemfile: 

- https://github.com/moneyadviceservice/comfortable-mexican-sofa/pull/24
- https://github.com/moneyadviceservice/mastalk/pull/23

What: The above description is related to the performance of the following cms requests:

- /en/articles/published 
- /:locale/:layout/:slug (e.g /en/articles/benefit-changes or en/corporate/corporate-news etc)

## N+1 Queries

There were some queries that was not optimised. 

Categories: It took 28 seconds to gather the categories for all the articles in the database. This was 8% of the total time of the request. On these 8% the ruby-prof shows above the total amount of time and the percentage of that 8% in each method.

cms-2

After the removal of N+1 queries, you can see the method category names is so fast that it took only miliseconds now:

category_names

Site and layout: This was not make too much difference but it takes 0.03% of the total and 0.14% of the total time. Add includes to make sure 

## Markdown processing

As you can see in the image below the markdown processing was taking 50% of the time to complete every request. This PR address the "cache" solution to solve the problem. Basically every time you save a page, cms will convert the markdown to html and on then serve the html on the request.

I opened a Mastalk PR to speed up the conversion too. You can see more details here: moneyadviceservice/mastalk#23

blocks

## Benchmarks

*Disclaimer*: The benchmarks should not be takens as the same numbers as production since thins benchmarks were runned as "development" mode, but they demonstrate the optimisations of this PR.

## Before the PR

⤷  ab -n 200 -c 15 http://localhost:3000/en/articles/the-best-ways-to-pay-bills
Concurrency Level:      15
Time taken for tests:   41.304 seconds
Complete requests:      200
Requests per second:    4.0 [#/sec] (mean)

## After the PR

⤷  ab -n 200 -c 15 http://localhost:3000/en/articles/the-best-ways-to-pay-bills
Concurrency Level:      15
Time taken for tests:   20.680 seconds
Complete requests:      200
Requests per second:    8.45 [#/sec] (mean)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -78,6 +78,7 @@ group :test, :development do
   gem 'valid_attribute', require: false
   gem 'dotenv-rails'
   gem 'launchy'
+  gem 'ruby-prof', require: false
 end
 
 group :production do
--- a/Gemfile
+++ b/Gemfile
@@ -12,7 +12,7 @@ gem 'coffee-rails', '~> 4.0.0'
 gem 'jquery-rails'
 gem 'jbuilder', '~> 2.0'
 
-gem 'comfortable_mexican_sofa', '~>2.0.1', path: '../comfortable-mexican-sofa'# git: 'git@github.com:moneyadviceservice/comfortable-mexican-sofa.git'
+gem 'comfortable_mexican_sofa', '2.0.1.79.79', source: 'http://gems.dev.mas.local'
 
 gem 'word-to-markdown', git: 'https://github.com/moneyadviceservice/word-to-markdown.git'
 
@@ -39,7 +39,7 @@ gem 'clockwork'
 gem 'paper_trail'
 gem 'feature'
 gem 'httparty', '~> 0.13.7'
-gem 'mastalk', '~> 0.8.0'
+gem 'mastalk', '0.9.0'
 gem 'mailjet'
 gem 'azure', '0.7.7'
 gem 'paperclip-azure', '~> 0.2'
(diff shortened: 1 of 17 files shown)


tomas-stefano:
The code climate issues is not related to this PR. I extract some methods from the classes to be reused and the Code Climate scan the whole files and complain about them. So the Code Climate issues will not be addressed on this PR.

Gemfile
```diff
@@ -12,7 +12,7 @@ gem 'coffee-rails', '~> 4.0.0'
 gem 'jquery-rails'
 gem 'jbuilder', '~> 2.0'
 
-gem 'comfortable_mexican_sofa', '~>2.0.1', git: 'git@github.com:moneyadviceservice/comfortable-mexican-sofa.git'
+gem 'comfortable_mexican_sofa', '~>2.0.1', path: '../comfortable-mexican-sofa'# git: 'git@github.com:moneyadviceservice/comfortable-mexican-sofa.git'
```
norm: This seems like a bad idea?

norm:
Can we get more context in the individual commits as to what and why plz :)

Gemfile
tomas-stefano: Actually, no. I will merge the comfy PR first and then address the proper version on the Gemfile, so this line could be ignored until I merge the PR and just link the version. I added before the explanation on the PR description about it,  perhaps I should have put more clearer.

tomas-stefano:
@norm good idea. The commit messages are very bad written. I will reword all of them and add the necessary explanations.

tomas-stefano:
@norm ♻️

norm:
Much better 👍

rnalexander:
:+1: :shipit:

PR #374, opened 2017-01-05
[DEV APPROVED] - 7832 - offline content editable

johnplayer1982:
# Adding offline page to CMS

This PR adds a Offline content page to the CMS

image


rnalexander:
:+1:

tomas-stefano:
:+1:

johnplayer1982:
This is no longer required, The information used on the offline content page is the same as the footer.  We should refactor the CMS to include a "contact details" page to put the phone number and opening hours in, as they will be used for the offline page and the footer

PR #375, opened 2017-01-10
Add migration to create extra empty clump links

jongilbraith:
Rather than add validation on the number of links a clump can have (which would have been a more complex UX addition) we went with creating a fixed number of empty links in the db and then showing them based on whether or not they were complete.

This approach wasn't finalised when the migration to create the initial clumps was created, so this adds the missing empty ones so that all clumps have four each.


tomas-stefano:
LGTM 👍

Tr4pSt3R:
all good 👍

jongilbraith:
@tomas-stefano @Tr4pSt3R thanks both

PR #377, opened 2017-01-12
Clump link changes to go with the new ‘Retirement’ clump

jongilbraith:
I rushed the PR to create the new Retirement clump and unfortunately forgot to consider the clump links.

The migration in this PR moves retirement related links over to the Retirement clump and then creates extra empty ones so that both the old and new clumps have four each.


tomas-stefano:
👍

Tr4pSt3R:
👍

PR #378, opened 2017-02-02
7915 - rework migrations and use final data ready for production

jongilbraith:
This PR takes all the previous migrations that were creating clumps, clumpings and links and replaces them with a minimal set that does everything in one go and to the state we expect to go live with.

All the clump copy has been confirmed by Ed and translated properly.

Additionally I took @jriga's script that created the new categories and put that in a migration so everything will be done in a single call to `rake db:migrate`.


tomas-stefano:
LGTM 👍

Tr4pSt3R:
Yup. it's all good 👍 👍

PR #381, opened 2017-02-09
Remove the “embedded google analytics” functionality

jongilbraith:
In the interest of getting the pipeline unblocked asap, releasing the removal of the GA integration in two steps.

This removes the GA stats display from the CMS interface, removing the js errors that are currently breaking the tests.

I will release soon (half way through) another PR to remove some more related functionality which involves a scheduled backend task that polls GA to update stats. It's quite far reaching and is probably best approached without time pressure.


tomas-stefano:
👍

Tr4pSt3R:
👍

norm:
Can we get the context added to the commit not the PR?

jongilbraith:
@norm the commit removes the functionality to display the the embedded GA stats and that's what the commit message says.

Should someone looking at the commit history always be given the entire reasoning and context for every change from every single commit?

I'm on the same page as you when it comes to good and informative commits but I don't see the necessity to go any further in this situation. While people should be expected to write good commit messages, people should also be expected RTF git log if needed to understand what might have happened in the past.

norm:
I like the why as well as the what in the commits. In my opinion, the extra context in the PR "This removes the GA stats display from the CMS interface, removing the js errors that are currently breaking the tests." would have been useful to add.

My rule of thumb has always been if I feel the need to add an explanation to the PR, I should probably have had that in the commit instead.

PR #383, opened 2017-03-01
[DEV APPROVED]7938 resize blog homepage images

Tr4pSt3R:
[View Card on Target Process](https://moneyadviceservice.tpondemand.com/entity/7938)

The MAS website is reported to have some performance and graphics quality issues owing to the size of the images that we use on the homepage for the blog thumbnail preview images. 

This PR provides a mechanism for generating a list of image _srcsets_ for the browser to load the appropriate image for devices of varying width.

homepage_thumbnails

**Dependent on this Frontend PR**: https://github.com/moneyadviceservice/frontend/pull/1691
**Dependent on this Comfortable Mexican Sofa PR**: https://github.com/moneyadviceservice/comfortable-mexican-sofa/pull/27

# Deployment Notes
Run ```rake paperclip:refresh:missing_styles``` to regenerate missing thumbnail styles for the `Comfy::Cms::File` class. This is the only class on the CMS project using the Paperclip attachment gem. 

Note: Ignore running refreshing the paperclip styles using the official Paperclip rake task. As an alternative, run the following code in the console instead. 

```Comfy::Cms::File.find_each(batch_size: 50) { |f| f.file.reprocess! }```

### CODE REVIEW

**Comment by norm** [comment]
This PR as it stands is a 👎 from me. Whilst the end result (as seen in the "Files changed" tab) looks good, the narrative that plays out in each commit is not. There's code that gets deleted and then reinstated (probably caused by merge collisions or bad rebasing), and several commits that have no good explanation.

I think it needs to be reworked so that it is one or two good commits. If you're not sure how, I'm happy to pair with you to make it better.

**Comment by rnalexander** [comment]
Please to not merge until have you've had a chance to pair with Norm on how to improve the structure.

File: app/views/files/_form.html.haml [hunk 1]
```diff
@@ -7,7 +7,10 @@
   = form.form_group t('.current_file') do
     - if @file.is_image?
       .file-image= image_tag @file.file.url
-    = link_to @file.file.url, @file.file.url, target: '_blank'
+    %ul
+      - FilesHelper::IMAGE_STYLES.each do |size|
+        %li
+          = link_to size.humanize, @file.file.url(size.to_sym), target: '_blank'
```
tomas-stefano: The file presenter above has the #full_path method. Can we use in this line?

spec/models/comfy/cms/file_spec.rb
```diff
@@ -0,0 +1,17 @@
+RSpec.describe Comfy::Cms::File do
+  describe '#styles' do
+    context 'various dimensions' do
+      {
+        extra_small: '390x244',
+              small: '485x304',
+             medium: '900x564',
+              large: '1350x846'
```
tomas-stefano: This line does not have a proper indentation with the first item on the hash. Can you run the rubocop on the files changed on the PR, please?

app/models/page_content_register.rb
```diff
@@ -17,7 +19,10 @@ def new_blocks_attributes
                         else
                           @new_blocks_attributes
                         end
-    return @new_blocks_attributes if home_page? || footer?
+
+    if home_page? || footer?
+      return blocks_attributes + srcset_blocks(blk_attrs: blocks_attributes)
```
tomas-stefano: Reading only this line, makes an assumption of "something is sum/adding things to the blocks attributes hash". But reading what the srcset_blocks method does, on my understanding, it is "changing the identifier prefix and add the styles of the block content of the block attributes if is raw_tile_\d_image". So, is this sum necessary? And do you think *srcset_blocks* is a good name to clarity on what the method is doing?

spec/models/page_content_register_spec.rb
```diff
@@ -30,9 +30,21 @@
...
+         { identifier: 'raw_tile_1_image', content: 'http://e.co/original/img.png' }]
+      end
 
-      it 'returns processed content as blank' do
-        expect(subject).to eq(new_blocks_attributes)
+      context 'image tile srcset' do
+        it 'is added to blocks attributes' do
+          blocks_identifiers = subject.map { |b| b[:identifier] }
+          expect(blocks_identifiers).to include('raw_tile_1_srcset')
+        end
```
tomas-stefano: I would change the context "when identifier is 'raw_title_\d_image'" and on the it I would change to " replaces identifier 'image'  with 'srcset'. Do you think is it more clear this way?

app/models/page_content_register.rb
```diff
@@ -26,6 +31,34 @@ def new_blocks_attributes
...
+  def available_styles
+    Comfy::Cms::File.new.file.styles
+      .reject { |k,_| [:cms_thumb, :cms_medium].include? k }
+      .map do |style, object|
+      {
+        style: style,
+        width: object.geometry.split('x').first
+      }
+    end
+  end
```
tomas-stefano: Is this file PageContentRegister the best location and the responsible for available_styles method? I would be honest that I don't know what the best place is, but it is worth we talk about it. And ... what would happen to the test ([here](https://github.com/moneyadviceservice/cms/pull/383/files#diff-9d0f2839f3bd8a7272090a62dc1bfc7dR46)) if you add more styles?

app/views/files/_form.html.haml
Tr4pSt3R: @tomas-stefano there's fix for this issue [here](https://github.com/moneyadviceservice/comfortable-mexican-sofa) Wdyt?

app/models/page_content_register.rb
Tr4pSt3R: I have had a look at it and played around with changing the code to eliminate the summation. Nonetheless, I can't figure out what's wrong with adding these two arrays; _block_attributes_ and _srcset_blocks_. If there is a better alternative I'm all ears to receive further feedback on this.

Tr4pSt3R: _srcset_blocks_ may not be the best so I will switch it to __new_srcset_blocks_.

app/models/page_content_register.rb
Tr4pSt3R: These tests will comfortably admit new style definitions as it's similar to the actual implementation. Does that answer your question on the specs or is there more to it that hasn't been covered. Let me know please.

Tr4pSt3R: The next best thing is probably to stick this in the libs directory or make it a model concern. Wdyt?

jriga: We could have something like `Comfy::Cms::File.available_styles`?

tomas-stefano: Yes, this sounds a good idea.

app/models/page_content_register.rb
tomas-stefano: Can you explain to me on what the srcset_blocks (or _new_srcset_blocks) does, please?

Tr4pSt3R: @tomas-stefano _new_srcset_blocks_ adds new block attributes which contain _srcset_ attributes for each image on the homepage. It's injected into the _new_blocks_attributes_ if they do not exist.

Tr4pSt3R:
♻️
@norm @rnalexander @tomas-stefano @jriga

tomas-stefano:
👍 👍

rnalexander:
@Tr4pSt3R Let me know when the last test we asked about is in place and I'll review it and thumb.  Thanks again for all the work on improving this PR!

Tr4pSt3R:
@rnalexander this PR is okay to be reviewed. It's this [Frontend PR](https://github.com/moneyadviceservice/frontend/pull/1691) that needs attention and a [WIP].

Tr4pSt3R:
@rnalexander @jriga @norm @pmcostello

jriga:
👍

rnalexander:
Thanks for the extra work on the PR on this!  :+1: :shipit:

PR #384, opened 2017-03-02
[DEV APPROVED]7995 - Floating Toolbar bug

johnplayer1982:
# Floating Toolbar Bug

**Description**

Currently the editor toolbar is floating over the content of the page rather than being fixed at the top, also the background isnt showing:

screen shot 2017-03-02 at 15 56 57

On further investigation I found that we are using the CSS property ``position: sticky`` and a fallback javascript plugin [filament-sticky](https://github.com/filamentgroup/fixed-sticky) AS this is still not supported by IE at all.  We are also using a custom dough component that initiates the plugin.

## The Solution

In order to make this as simple as possible I decided to remove the 3rd party plugin (As this seems to have stopped working anyway) and update the component to do to work of applying a position: fixed property, rather than trying to use position: sticky, then trying to patch in a fix for unsupported browsers.

To give a bit of perspective of the support for each of the CSS properties:

| position: sticky; |
|-----------------------|
|screen shot 2017-03-03 at 09 00 46|

| position: fixed; |
|---------------|
|screen shot 2017-03-03 at 09 00 10|

## The Outcome

screen shot 2017-03-03 at 09 15 27

### DIFF
--- a/app/assets/javascripts/comfortable_mexican_sofa/admin/modules/StickyElements.js
+++ b/app/assets/javascripts/comfortable_mexican_sofa/admin/modules/StickyElements.js
@@ -1,34 +1,48 @@
-define([
-  'jquery',
-  'DoughBaseComponent',
-  'filament-sticky'
-], function (
+define([ 'jquery', 'DoughBaseComponent'], function (
   $,
   DoughBaseComponent
-) {
+  ) {
   'use strict';
 
   var StickyElementsProto,
-      defaultConfig = {
-        selectors: {
-          stickyElement: '[data-dough-sticky-element-target]'
-        }
-      };
+  defaultConfig = {
+    selectors: {
+      stickyElement: '[data-dough-sticky-element-target]'
+    }
+  };
 
   function StickyElements($el, config) {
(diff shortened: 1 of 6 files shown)


davidtrussler:
Haven't been able to test locally properly but 👍 looks good

thecountgs:
👍 I was able to test the CMS locally and toolbar looks good.

tomas-stefano:
It is not related to the problems but we have references for the filament stick plugin on the code:

```
app/assets/stylesheets/comfortable_mexican_sofa/admin/helpers/_fixedsticky.scss
1:// Helper for the filament-sticky plugin

config/initializers/assets.rb
56:                                  filament-sticky/fixedsticky.js
57:                                  filament-fixed/fixedfixed.js
```

Do they need to be deleted too?

johnplayer1982:
The CSS is still used as it still describes the re-written component, but the 2 lines in config/initializers/assets.rb do need to be removed, apologies for missing it, ill create a PR now

johnplayer1982:
PR to remove these: https://github.com/moneyadviceservice/cms/pull/385

PR #388, opened 2017-04-03
[WIP] Load Universal Credit advices and tips in the CMS

jriga:
**Purpose**
Extract advices and tips content in `universal_credit-content` to english and welsh pages in the CMS.

**Proposed Solution**
Added a rake tasks to:
 - `bundle exec rake layout:create[universal_credit]` 
    creates english and welsh layouts for universal_credit

 - `bundle exec rake uc:add_pages[vendor/universal_credit/en]`
    creates english and welsh pages 

**Devs**
@tomas-stefano, @jriga

PR #390, opened 2017-04-10
changes page slug to uc content title

Tr4pSt3R:
Adds support for loading content from the CMS. 

See https://github.com/moneyadviceservice/cms/pull/388


jriga:
👍

PR #395, opened 2017-07-03
[DEV APPROVED] -8198 - Set published date for scheduled posts

shaley91:
### Summary
This pr sets a published date of a page to be when its scheduled to be released.

The CMS app is not revising the published_at date on articles with scheduled revisions upon publication, although the content is updated, the published_at date is required by the web feeds management system to update partners on update articles in their feeds.

[TP link](https://moneyadviceservice.tpondemand.com/restui/board.aspx?acid=13857292912bd09b911bf771e091f219#page=bug/8198)

### DIFF
--- a/spec/models/comfy/cms/page_spec.rb
+++ b/spec/models/comfy/cms/page_spec.rb
@@ -251,6 +251,24 @@ RSpec.describe Comfy::Cms::Page do
     end
   end
 
+  describe '.scheduled_today' do
+    context 'pages scheduled in the past, present and future' do
+      it 'includes only pages scheduled for today' do
+        page_one = create :page, state: :scheduled, scheduled_on: Time.now.end_of_day - 2.hours
+        page_two = create :page, state: :scheduled, scheduled_on: Time.now
+        create :page, state: :scheduled, scheduled_on: Time.now.end_of_day + 1.day
+        create :page
+        create :page
+        create :page
+
+        expect(Comfy::Cms::Page.scheduled_today).to contain_exactly(
+          page_one,
+          page_two
+        )
+      end
+    end
+  end
+
   describe '#most_popular scope' do
 
     it 'has the three most popular articles' do
--- a/app/models/comfy/cms/page.rb
+++ b/app/models/comfy/cms/page.rb
@@ -48,6 +48,13 @@ class Comfy::Cms::Page < ActiveRecord::Base
 
   scope :unpublished, -> { where(state: 'unpublished') }
 
+  scope :scheduled_today, (lambda do
+    where(
+      arel_table[:state].eq('scheduled')
+      .and(arel_table[:scheduled_on].lt(Time.now.end_of_day))
+    )
+  end)
+
   scope :layout_identifier, (lambda do |identifier|
     joins(:layout).where(comfy_cms_layouts: { identifier: identifier.singularize })
   end)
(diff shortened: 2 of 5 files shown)


tomas-stefano:
There is a code climate issue. Can you verify, please?

shaley91:
@tomas-stefano it thinks that `save_post` is now a complex method.  I don't think its anything to worry about.

tomas-stefano:
I would follow the code climate issue and try to think a best place rather than a controller to do this. Do the registers classes has a parent ancestor in common?

shaley91:
@tomas-stefano we are finding it hard to work out a better place to put this code?  It should really live in the same place that marks scheduled posts as `published` but we have no idea where that happens.  Any help would be much appreciated.

tomas-stefano:
A suggestion will be to create an object and move this save page controller method to there? IMHO, the pages controller is not the right place to have all steps/rules on how to save a page. Is that makes sense?

shaley91:
@tomas-stefano - On further searching it appears that the issue we are trying to solve is because scheduled posts in the cms never actually get published.

https://github.com/moneyadviceservice/comfortable-mexican-sofa/blob/master/app/models/comfy/cms/page.rb#L75

It would appear that the correct fix is to publish the the page rather then rely on this catch.  Normally I would write a background job, but I don't think the cms has background jobs.  Is this correct?

Given that we don't have a background job, where is the best place to publish a scheduled post? In an after commit block? Or a presenter? Or a controller?

tomas-stefano:
No the CMS does have some sort of background jobs (see lib/clock.rb) but it is not what you're looking for. IMHO, I think it is not the best approach to use background jobs to rely on schedule, otherwise you'll need to change in a lot of places (how the publish page is retrieved and how the page is created).

I would create a ruby class for that (e.g PageCreator) that will hold some of the rules of creating the page. You can start by moving the save_page controller method logic to this class. And then everything news related to the Page creator goes to this class. Wdyt?

shaley91:
I don't think we need a PageCreator class.  There is already a lot of things that edit pages, and they are spread out all over this app.  Creating another place just makes things more messy, unless you want to tidy everything up? (which is definitely not in the scope of this ticket).

The code we wrote in this ticket is wrong and should be removed.  The only thing that needs to happen to fix the documented bug is for a page to be published if the scheduled time is in the past.  I just don't know where to put that line.

shaley91:
This PR now uses a job to set the published at time of pages that are scheduled to be published in the next day.

We did this because:
* We don't want to set all pages published at because it might cause unexpected results due to things checking a published at timestamp. This is normally only set when a page is published.
* Transitioning a page from scheduled to publish is currently not supported in the MAS/ComfyMexicanSofa fork.

tomas-stefano:
👍

spec/models/comfy/cms/page_spec.rb
```diff
@@ -251,6 +251,22 @@
     end
   end
 
+  describe '#scheduled_today scope' do
```
Tr4pSt3R: Great tests but the structure of the test can be re-organised to communicate the true intention of the test. 

```
describe '#scheduled_today' do
  context 'pages scheduled in the past' do
    it 'includes only pages scheduled for today' do
       ...
    end
  end
end
```

tomas-stefano: Good spot. And to add more, this is a class method so it should be ".scheduled_today" as a convention.

PR #396, opened 2017-09-27
[DEV APPROVED] 8603 Disable category from being selected as the parent

margOnline:
### Objective
This pr addresses [TP8603](https://moneyadviceservice.tpondemand.com/restui/board.aspx?acid=ddd7f3a626844ab8d1bee3f06ec28efa#page=board/5624876525867731357&appConfig=eyJhY2lkIjoiREI4ODcwRjkxMDNDRTM2NTlBMzhDNTRBRTVBNUU1N0UifQ==&boardPopup=bug/8603).  A user design flaw in production allows a content editor or admin user to assign a category as the parent category to itself. If the user then tries to view that category page, he'll experience a timeout. Technically, the cause is an infinite loop in the code.

The short term approach is to disable a category from being selected as its own parent in the drop down list on the form for editing categories. A better approach is to validate the parent category when the form is submitted. This latter approach needs to be prioritised. In the interim, the solution of disabling the parent category from being selected should prevent an immediate recurrence of the error.

### Technical
Cucumber scenario to specifically test that a category is not able to be selected as its parent category:

* New feature for editing a category specifically to change the parent category
* To populate the parent categories dropdown menu, categories were being selected on the basis of the site id equalling 1. Consequently, the test database needs to be reset, otherwise the tests will fail after the first time since the site id will continually be incremented.
* Tagged the scenario for use with an after hook so the database cleaner will only kick in for this test. Not sure if this is necessary but added it so the entire test suite is not slowed down.

### DIFF
--- a/features/support/ui/pages/category_details.rb
+++ b/features/support/ui/pages/category_details.rb
@@ -1,0 +1,11 @@
+require_relative '../page'
+require_relative '../sections/header'
+
+module UI::Pages
+  class CategoryDetails < UI::Page
+    set_url         '/admin/categories/{id}'
+    set_url_matcher /\/admin\/categories\/\d+/
+
+    elements :parent_options, '#comfy_cms_category_parent_id option'
+  end
+end
--- a/app/controllers/categories_controller.rb
+++ b/app/controllers/categories_controller.rb
@@ -8,6 +8,7 @@ class CategoriesController < Comfy::Admin::Cms::BaseController
   def show
     @category.links.build
     @category.category_promos.build
+    @categories = Comfy::Cms::Category.where(site_id: 1).reorder(:label)
     @english_pages, @welsh_pages =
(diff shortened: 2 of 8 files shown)


features/support/ui/pages/category_details.rb
```diff
@@ -0,0 +1,12 @@
+require_relative '../page'
+require_relative '../sections/header'
+
+module UI::Pages
+  class CategoryDetails < UI::Page
+    set_url         '/admin/categories/{id}'
+    set_url_matcher /\/admin\/categories\/\d+/
+
+    elements :parent_options, '#comfy_cms_category_parent_id option'
+
```
Tr4pSt3R: this line can go :)

tomas-stefano:
👍 (after the minor thing that Jones posted)

PR #399, opened 2018-02-08
[DEV APPROVED] Add fincap rake task to setup database

tomas-stefano:
[TP](https://moneyadviceservice.tpondemand.com/entity/8917-create-a-database-setup-for-evhub)

The purpose of this PR is to have a database task which we can run in development and in production on the first **Fincap CMS** deploy and enable the editors to start creating an Insight page. Note: the user and password will change but we can do on CMS interface itself.

This can run as:`rake fincap:setup`

This solves the following problem:
Given I am an editor
When I try to create a new page
Then I should see an option to have an Insights layout


lib/tasks/fincap.rake
```diff
@@ -0,0 +1,25 @@
+namespace :fincap do
```
Tr4pSt3R: Could we not have used `db/seed.rb` to seed these files?

tomas-stefano: The cms db/seeds.rb is used mainly for development but this rake task should be run in production as well so we choose a rake task because then the code will not have data creation altogether CMS and Fincap.

Tr4pSt3R: It's unconventional to create data using a rake task especially when you can seed production database with `RAILS_ENV=production bundle exec rails db:seed`. Anyway if you are convinced this is the best approach then go for it.

tomas-stefano: Sorry, perhaps I was not clear. The cms db/seeds.rb has a lot of data from our **current CMS**. **The Fincap CMS will not have the same data**, so run the seeds as you described will not solve the problem in any way I can see it. So that's the reason that we didn't added to the seeds nor any other method than a rake task. Please, let's talk in person or via other method for clarification on this.

PR #402, opened 2018-02-19
[DEV APPROVED] 8913 add text fields

willhall88:
[TP8913](https://moneyadviceservice.tpondemand.com/entity/8913-add-text-blocks-to-cms-insights)

Add insight template
Add cucumber feature
Add refactored code for insight partial
Add fields to template!

**refactoring form builder**
The method `default_tag_field` needs to be refactored to be maintainable. Although we have updated some of this method to work a little better we have opted not to refactor this whole method yet. This is down to the fact that we know we will have to do more work to this method with the addition of multiple blocks (checkboxes etc.). It may be unwise to refactor this code before that point as it may make it harder to then apply any new changes from that point. _Therefore_ we will address the refactor of this method in that PR
https://github.com/moneyadviceservice/cms/pull/402/files#diff-0b23ec871fc4b345b5c142d0d9d33424R45

### DIFF
--- a/app/views/pages/form_blocks/_insight.html.haml
+++ b/app/views/pages/form_blocks/_insight.html.haml
@@ -1,0 +1,3 @@
+%div
+  - fields_from_layout(namespace).each_with_index do |layout_field, index|
+    = layout_field.input_tag_for(form_builder: cms_blocks, index: index)
--- a/features/step_definitions/fincap_pages_steps.rb
+++ b/features/step_definitions/fincap_pages_steps.rb
@@ -1,0 +1,67 @@
+Given(/^I have an insight page layout$/) do
+  english_site = Comfy::Cms::Site.find_or_create_by(
+    label: 'en',
+    identifier: 'money-advice-service-en',
+    hostname: 'localhost:3000',
+    path: 'en',
+    locale: 'en',
+    is_mirrored: true
+  )
+
+  english_site.layouts.find_or_create_by(
+    identifier: 'insight',
+    label: 'Insight',
+    content:  <<-CONTENT
+      {{ cms:page:content:rich_text }}
+      {{ cms:page:overview }}
+      {{ cms:page:countries }}
+      {{ cms:page:links_to_research }}
+      {{ cms:page:contact_details }}
+      {{ cms:page:year_of_publication }}
+    CONTENT
+  )
+end
+
+Given(/^I am on the homepage$/) do
+  home_page.load
+end
+
+Given(/^I click on "([^"]*)"$/) do |button|
+  click_on(button)
+end
+
+Given(/^I select "([^"]*)" on the creation page$/) do |option|
+  new_page.select(option)
+end
+
+Given(/^I create a page "([^"]*)"$/) do |title|
+  new_page.page_name.set(title)
+  new_page.save_button.click
+end
+
+When(/^I edit the page "([^"]*)"$/) do |title|
+  home_page.load
+  page.first(:link, title).click
+end
+
+When(/^I fill in$/) do |table|  # table is a Cucumber::Core::Ast::DataTable
+  table.rows.each do |row|
+    edit_page.send(row[0]).set(row[1])
+  end
+end
+
+When(/^I save and return to the homepage$/) do
+  edit_page.save_changes_to_draft.click
+  home_page.load
+end
+
+When(/^when I click the "([^"]*)" page$/) do |title|
+  step(%(I edit the page "#{title}"))
+end
+
+Then(/^I should see the fields filled with the content$/) do |table|
+  # table is a Cucumber::Core::Ast::DataTable
+  table.rows.each do |row|
+    expect(edit_page.send(row[0]).value).to eq(row[1])
+  end
+end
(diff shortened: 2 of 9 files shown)


features/step_definitions/fincap_pages_steps.rb
```diff
@@ -0,0 +1,67 @@
...
+    CONTENT
+  )
+end
+
+Given(/^I am on the homepage$/) do
+  home_page.load
+end
+
+Given(/^I click on "([^"]*)"$/) do |button|
+  page.click_on(button)
```
Tr4pSt3R:
```ruby
click_on(button)
```

Should work just as well. It will not be in conflict with `new_page`. Once you've navigated to the right page calling the `page` object  itself is redundant.

features/step_definitions/fincap_pages_steps.rb
```diff
@@ -0,0 +1,67 @@
...
+end
+
+Given(/^I create a page "([^"]*)"$/) do |title|
+  new_page.page_name.set(title)
+  new_page.save_button.click
+end
+
+When(/^I edit the page "([^"]*)"$/) do |title|
+  home_page.load
+  page.first(:link, title).click
```
Tr4pSt3R: Is there not a specific identifier for the edit button? There must be a better way of finding a clickable object within the page. This element can be set in the page object class and used just as is in previous parts of this PR i.e. `page.edit_button.click`

What will happen if the title is changed in the slightest?

spec/lib/cms/layout_field_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+    let(:input_tag_for) do
+      layout_field.input_tag_for(form_builder: form_builder, index: 0)
+    end
+
+    context 'when comfortable mexican sofa tag is rich text' do
+      let(:comfortable_mexican_sofa_tag) { ComfortableMexicanSofa::Tag::PageRichText.new }
+      let(:form_builder) { double }
+
+      it 'renders page rich text input tag from the form builder' do
+        expect(form_builder).to receive(:page_rich_text).with(comfortable_mexican_sofa_tag, 0)
```
Tr4pSt3R: Can you add a single line to separate the spy from the implementation please?

app/views/pages/form_blocks/_insight.html.haml
```diff
@@ -0,0 +1,3 @@
+%div
+  - fields_from_layout(namespace).each_with_index do |layout_field, index|
+    = layout_field.input_tag_for(form_builder: cms_blocks, index: index)
```
margOnline: new line?

lib/cms/form_builder.rb
```diff
@@ -41,7 +42,7 @@ def default_tag_field(tag, index, method = :text_field_tag, options = {})
 
   # ACHTUNG, HACKY! - looks for the instance variable @blocks_attributes in the view to retrieve the current content
   blocks_attributes = @template.instance_variable_get('@blocks_attributes')
-  current_value = blocks_attributes.find { |block_attributes| block_attributes[:identifier] == tag.identifier.to_s }[:content]
+  current_value = Array(blocks_attributes).find { |block_attributes| block_attributes[:identifier] == tag.identifier.to_s }.try(:[], :content)
```
margOnline: line length?

features/step_definitions/fincap_pages_steps.rb
willhall88: not sure of the best solution to this. On the homepage there are two identical links to edit the page. Neither of the links have any specific identifier and the parents of them a fairly generic as well. The benefit of this solution is that *if* the title in the feature is changed slightly then it won't find the page, which should happen. We want to ensure this step is selecting the correct page to edit and the best way of ensuring this is to be using the title of the page, don't you think?

willhall88:
@margOnline @Tr4pSt3R - comments addressed, let me know your thoughts @Tr4pSt3R on my comment above ☝️

features/step_definitions/fincap_pages_steps.rb
Tr4pSt3R: @willhall88 If that's the case then maybe this solution is safe to toe the same line.

features/step_definitions/fincap_pages_steps.rb
```diff
@@ -0,0 +1,67 @@
+Given(/^I have an insight page layout$/) do
+  english_site = Comfy::Cms::Site.find_or_create_by(
```
Tr4pSt3R: We already have factories for creating the English site. I anticipate that this will change in the future. 
It's fine to have it here, as it is, for now though.

features/step_definitions/fincap_pages_steps.rb
```diff
@@ -0,0 +1,67 @@
...
+    label: 'en',
+    identifier: 'money-advice-service-en',
+    hostname: 'localhost:3000',
+    path: 'en',
+    locale: 'en',
+    is_mirrored: true
+  )
+
+  english_site.layouts.find_or_create_by(
```
Tr4pSt3R: The same thing with the layout, we already have a layout factory. We can add traits for this later on to DRY our implementation as the code grows to accommodate Evaluation and Review layout types.
Again it's fine to have the basics here as you've done.
It will change in the near future though.

willhall88:
♻️ rebased with fincap branch

PR #405, opened 2018-03-01
[DEV APPROVED] 8997 API Search by keyword - Insight

margOnline:
[TP 8997](https://moneyadviceservice.tpondemand.com/restui/board.aspx?#page=board/5580782241563718174&appConfig=eyJhY2lkIjoiRTQ4MDEyNENDOUYxRDU4RDg0RTEyRjNCODBEN0VBMUYifQ==&searchPopup=userstory/8997)

Uses the `keyword` parameter from the request to search page content and return an array of documents which contain the `keyword`

### Technical
The [Comfortable Mexican Sofa gem](https://github.com/moneyadviceservice/comfortable-mexican-sofa) provides scopes which the api can use for cleaner looking queries. We've used the [with_content_like](https://github.com/moneyadviceservice/comfortable-mexican-sofa/blob/master/app/models/comfy/cms/page.rb#L104) scope to implement the keyword search.

### Next Steps
1. Search title
2. Search overview

### DIFF
--- a/app/controllers/api/documents_controller.rb
+++ b/app/controllers/api/documents_controller.rb
@@ -16,11 +16,20 @@ module API
       params[:document_type]
     end
 
+    def keyword
+      params[:keyword]
+    end
+
     def filter_documents
       if document_type.present?
         @documents = @documents.joins(:layout)
           .where('comfy_cms_layouts.identifier' => document_type)
       end
+
+      if keyword.present?
+        @documents = @documents.with_content_like(keyword)
+      end
     end
+
   end
 end
--- a/spec/controllers/api/documents_controller_spec.rb
+++ b/spec/controllers/api/documents_controller_spec.rb
@@ -6,38 +6,71 @@ RSpec.describe API::DocumentsController, type: :request do
   let(:response_body) { JSON.load(response.body) }
   let(:documents) { response_body['documents'] }
   let(:meta_data) { response_body['meta'] }
-  let(:url) { "/api/en/documents#{params}" }
+  let(:url) { '/api/en/documents' }
+  let(:review_layout)  { create :layout, identifier: 'review' }
+  let(:insight_layout)  { create :layout, identifier: 'insight' }
+  let!(:insight_page) { create(:insight_page_about_financial_wellbeing, site: site, layout: insight_layout) }
+  let!(:insight_page_1) { create(:insight_page_about_debt, site: site, layout: insight_layout) }
+  let!(:insight_page_2) { create(:insight_page_about_pensions, site: site, layout: insight_layout) }
+  let!(:review_page_1) { create(:page, site: site, layout: review_layout) }
 
   before do
     allow_any_instance_of(PageSerializer)
       .to receive(:related_content).and_return({})
+
+    get url, params
   end
 
   describe 'GET /:locale/documents' do
-    let(:review_layout)  { create :layout, identifier: 'review' }
-    let(:insight_layout)  { create :layout, identifier: 'insight' }
-
-    let!(:insight_page_1) { create(:page, site: site, layout: insight_layout) }
-    let!(:insight_page_2) { create(:page, site: site, layout: insight_layout) }
-    let!(:review_page_1) { create(:page, site: site, layout: review_layout) }
-
-    before { get url }
-
     context 'when all documents are requested' do
-      let(:params) {''}
+      let(:params) { {} }
 
       it 'returns all documents' do
-        expect(meta_data['results']).to eq 3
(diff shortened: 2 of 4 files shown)


spec/controllers/api/documents_controller_spec.rb
```diff
@@ -7,37 +7,67 @@
...
+  let!(:insight_page) { create(:insight_page_about_financial_wellbeing, site: site, layout: insight_layout) }
+  let!(:insight_page_1) { create(:insight_page_about_debt, site: site, layout: insight_layout) }
+  let!(:insight_page_2) { create(:insight_page_about_pensions, site: site, layout: insight_layout) }
+  let!(:review_page_1) { create(:page, site: site, layout: review_layout) }
 
   before do
     allow_any_instance_of(PageSerializer)
       .to receive(:related_content).and_return({})
+
+    get url
```
Tr4pSt3R: @margOnline I think we can do something like this:

```ruby
let(:url) { '/api/en/documents' }

before do
  ...
  get url, params: params
```

I will verify this locally. This is a much better way though. Wdyt?

tomas-stefano:
@margOnline which point is best to benchmark and profiling this search? I am asking because in a simple way we are doing a sort of full text search through entire database. But I don't want to premature optimisation either.

app/controllers/api/documents_controller.rb
```diff
@@ -16,11 +16,20 @@ def document_type
...
+
     def filter_documents
       if document_type.present?
         @documents = @documents.joins(:layout)
           .where('comfy_cms_layouts.identifier' => document_type)
       end
+
+      if keyword.present?
+        @documents = @documents.with_content_like(keyword)
+      end
```
tomas-stefano: When is the best time to move this logic out of the controller to an object (DocumentFilter.filter perhaps)?

spec/controllers/api/documents_controller_spec.rb
margOnline: In order to take this approach, the params need to be defined as hashes in the individual context blocks which is fine for something like 
`let(:params) { {document_type: 'insight', keyword: 'pension' } }` 
but for a phrase like 'Financial well being: the employee view', it has to be passed like as

```
{
  keyword: 'Financial well being: the employee view'
}

 whereas in the url it would be passed as: `Financial+well+being%3A+the+employee+view` so the former nicely tests that the above param is correctly converted to 'Financial well being: the employee view'

File: app/controllers/api/documents_controller.rb (hunk 2 above)
Comment (margOnline) [reply app/controllers/api/documents_controller.rb]: Not sure, but not at this time.

**Comment by margOnline** [comment]
Re performance/benchmarking, I agree with the above comment. I will add it in when I enhance the query for title or overview search

File: app/controllers/api/documents_controller.rb [hunk 3]
```diff
@@ -16,11 +16,20 @@ def document_type
...
+    end
+
     def filter_documents
       if document_type.present?
         @documents = @documents.joins(:layout)
           .where('comfy_cms_layouts.identifier' => document_type)
       end
+
+      if keyword.present?
+        @documents = @documents.with_content_like(keyword)
```
paulanthonywilson: I'm not sure what kind of load Fincap is looking at supporting. It might be worth considering load- testing under load, with a full database.

spec/controllers/api/documents_controller_spec.rb
Tr4pSt3R: @margOnline you will eventually have to define _params_ in the _context_ for every example block as a hash rather than as a text.

margOnline: regards @Tr4pSt3R's point, wdyt? @willhall88 @tomas-stefano

tomas-stefano: I am not sure. I need to checked the code and test @Tr4pSt3R option to see if is better.

tomas-stefano: I tested and Jones point makes sense. 

The code becomes more cleaner and the app parses right:

```
1890:Started GET "/api/en/documents?document_type=insight&keyword=Financial+well+being%3A+the+employee+view" for 127.0.0.1 at 2018-03-02 12:21:59 +0000
1892-Processing by API::DocumentsController#index as HTML
1893-  Parameters: {"document_type"=>"insight", "keyword"=>"Financial well being: the employee view", "locale"=>"en"}
```

The branch that I did the testing https://github.com/moneyadviceservice/cms/compare/8997-insight-keyword-search...testing-jones-idea

margOnline: 👍

PR #406, opened 2018-03-02
Document api with Apipie

margOnline:
This pr explores one option for documenting the cms api, namely using [Apipie](https://github.com/Apipie/apipie-rails).

Documentation for some of the api methods has been included as an example. This approach is open for discussion and is merely a response to a request to look into how we document the cms api.

### DIFF
--- a/app/controllers/api/category_contents_controller.rb
+++ b/app/controllers/api/category_contents_controller.rb
@@ -5,11 +5,16 @@ module API
     skip_before_action :load_cms_page
     around_filter :validate_locale
 
+    api :GET, '/:locale/categories'
+    param :locale, /[en|cy]/, required: true
     def index
       @primary_navigation, @secondary_navigation = Comfy::Cms::Category.navigation_categories
       render json: @primary_navigation, scope: locale
     end
 
+    api :GET, '/:locale/categories/:id'
+    param :locale, /[en|cy]/, required: true
+    param :id, :number, required: true
     def show
       @category = Comfy::Cms::Category.find_by(label: params[:id])
       render json: @category, scope: locale
--- a/config/initializers/apipie.rb
+++ b/config/initializers/apipie.rb
@@ -1,0 +1,8 @@
+Apipie.configure do |config|
+  config.app_name                = "Cms"
+  config.api_base_url            = "/api"
+  config.doc_base_url            = "/apipie"
+  config.api_controllers_matcher = "#{Rails.root}/app/controllers/**/*.rb"
+  config.default_locale          = 'en'
+  config.languages               = ['en']
+end
--- a/Gemfile
+++ b/Gemfile
@@ -2,6 +2,7 @@ source 'http://gems.dev.mas.local'
 source 'https://rubygems.org'
 
 gem 'active_model_serializers'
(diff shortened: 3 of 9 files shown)


config/initializers/apipie.rb
```diff
@@ -0,0 +1,8 @@
+Apipie.configure do |config|
+  config.app_name                = "Cms"
+  config.api_base_url            = "/api"
+  config.doc_base_url            = "/apipie"
+  config.api_controllers_matcher = "#{Rails.root}/app/controllers/**/*.rb"
```
Tr4pSt3R: Should this be pointing to the`ApiController`?

app/controllers/api/category_contents_controller.rb
```diff
@@ -5,11 +5,18 @@ class CategoryContentsController < ApplicationController
     skip_before_action :load_cms_page
     around_filter :validate_locale
 
+    api :GET, '/:locale/categories'
+    param :locale, /[en|cy]/, required: true
+    param :page_type, String, required: true
```
tomas-stefano: Does categories has page_type?

app/controllers/api/category_contents_controller.rb
```diff
@@ -5,11 +5,18 @@ class CategoryContentsController < ApplicationController
...
     def index
       @primary_navigation, @secondary_navigation = Comfy::Cms::Category.navigation_categories
       render json: @primary_navigation, scope: locale
     end
 
+    api :GET, '/:locale/categories/:id'
+    param :locale, /[en|cy]/, required: true
+    param :id, :number, required: true
+    param :page_type, String, required: true
```
tomas-stefano: Does this route has page_type?

margOnline:
@tomas-stefano @Tr4pSt3R @willhall88 the purpose of the pr was to open the discussion into how/if we document the cms api which [Tomas raised several times in the slack channel]( https://moneyadviceservice.slack.com/archives/C33B2BM45/p1518690637000014). It's probably not worth implementing the code for all the api routes if we're not going to use apipie.

Initially could you provide feedback on apipie itself and/or any api documentation approaches you prefer e.g. do you think apipie is worth implementing, is it easy to use, does it provide the kind of documentation you would expect/want from an api, would you use it, are you happy with the additional dsl that would be required in the files? or you're not bothered what is used.

config/initializers/apipie.rb
margOnline: no, because we want different controllers to be picked up

tomas-stefano:
@margOnline sounds good to me this PR to be merged to just adding the Documents API doc.

Then we use the following parameters:
* api
* short
* param
* error
* description (explaining with some example for some parameters).

e.g

```
  api :GET, "/api/en/documents", "List all documents from CMS"
  short "..."
  param :document_type, Array, desc: '...'
  param :keyword, String, desc: '...'
  param :blocks, Hash, :desc => 'Filter by identifier and content' do
    param :identifier, String, :desc => "Username for login", :required => true
    param :content, String, :desc => "Password for login", :required => true
  end
  error 204, "No documents found"
  description <<-EOS
     == Examples
          GET /api/en/documents.json
     ...
  EOS
```

Wdyt?

PR #410, opened 2018-03-06
8997 API: Search page titles for search term

margOnline:
[TP 8997](https://moneyadviceservice.tpondemand.com/restui/board.aspx?#page=board/5580782241563718174&appConfig=eyJhY2lkIjoiRTQ4MDEyNENDOUYxRDU4RDg0RTEyRjNCODBEN0VBMUYifQ==&searchPopup=userstory/8997)

This pr adds searching through page titles for a given keyword.

### Technical
The query needs to return pages where the keyword is in the title OR the content. Since chaining scopes only works for 'AND' search queries, it is necessary to use the more wordy `where` syntax.

### Next Steps
1. Search only blocks which are 'content' or 'overview'
2. Refactor the documents_controller#filter_documents method

### DIFF
--- a/spec/controllers/api/documents_controller_spec.rb
+++ b/spec/controllers/api/documents_controller_spec.rb
@@ -7,12 +7,16 @@ RSpec.describe API::DocumentsController, type: :request do
   let(:documents) { response_body['documents'] }
   let(:meta_data) { response_body['meta'] }
   let(:url) { '/api/en/documents' }
+  let(:insight_page_params) { {site: site, layout: insight_layout } }
   let(:review_layout)  { create :layout, identifier: 'review' }
-  let(:insight_layout)  { create :layout, identifier: 'insight' }
-  let!(:insight_page) { create(:insight_page_about_financial_wellbeing, site: site, layout: insight_layout) }
-  let!(:insight_page_1) { create(:insight_page_about_debt, site: site, layout: insight_layout) }
-  let!(:insight_page_2) { create(:insight_page_about_pensions, site: site, layout: insight_layout) }
-  let!(:review_page_1) { create(:page, site: site, layout: review_layout) }
+  let(:insight_layout) { create :layout, identifier: 'insight' }
+  let!(:insight_page1) { create(:insight_page_about_financial_wellbeing, insight_page_params) }
+  let!(:insight_page2) { create(:insight_page_about_debt, insight_page_params) }
+  let!(:insight_page3) { create(:insight_page_about_pensions, insight_page_params) }
+  let!(:insight_page4) { create(:insight_page_titled_annuity, insight_page_params) }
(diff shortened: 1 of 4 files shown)


spec/controllers/api/documents_controller_spec.rb
```diff
@@ -26,28 +28,39 @@
...
+          end
+        end
+
+        context 'and the keyword is found in the title' do
+          let(:params) { { document_type: 'insight', keyword: 'annuity' } }
+
+          it 'returns an array of documents which contain the keyword' do
+            expect(meta_data['results']).to eq 2
+            expect(documents.count).to eq 2
+          end
```
tomas-stefano: Can we assert about the document info instead of using documents.count? For the page type makes sense but for search by keyword would be good to assert that the document that you search is the document that is returned by the API.

spec/controllers/api/documents_controller_spec.rb
```diff
@@ -26,28 +28,39 @@
...
+        context 'and the keyword is found in the content' do
+          let(:params) { { document_type: 'insight', keyword: 'pension' } }
 
-        it 'returns an array of documents which contain the keyword' do
-          expect(meta_data['results']).to eq 1
-          expect(documents.count).to eq 1
+          it 'returns an array of documents which contain the keyword' do
+            expect(meta_data['results']).to eq 1
+            expect(documents.count).to eq 1
+          end
```
tomas-stefano: The same about the document assertion comment.

Gemfile.lock
```diff
@@ -7,12 +7,13 @@ GIT
 
 GIT
   remote: https://github.com/moneyadviceservice/dough.git
-  revision: cf4e4cbd5ee86ef49b2db16fc44dde1ed9f2c0f4
+  revision: f696d53ddac5f73d64a5d32ecbe03fd9d0b9298e
   specs:
-    dough-ruby (5.7.0)
+    dough-ruby (5.24.1)
```
willhall88: had to update dough-ruby again as it couldn't find this version.
latest is `5.27`

spec/controllers/api/documents_controller_spec.rb
```diff
@@ -26,28 +30,65 @@
...
+        end
+
+        context 'when the keyword is in the title of one document and content of another' do
+          let(:params) { { document_type: 'insight', keyword: 'annuity' } }
+
+          it 'returns both documents' do
+            documents.each do |doc|
+              doc_title = doc['label']
+              doc_content = doc['blocks'].select do |block|
+                block['identifier'] =='content' && block['content'].include?(params[:keyword])
```
willhall88: as an alternative to my suggestion above, if you change this line to just be
`block['identifier'] =='content'` as you don't need to check for the keyword in here, that is the responsibility of the search and thus the expectation below.

PR #412, opened 2018-03-12
modify query to select distinct pages

margOnline:
Fix for a bug in the search which was returning a page twice if the keyword was found in the title and the content block. By adding the `uniq` qualify we ensure that a page is only returned once.


johnplayer1982:
:+1:

PR #413, opened 2018-03-13
TP 8999 api search filtering

margOnline:
[TP 8999](https://moneyadviceservice.tpondemand.com/restui/board.aspx?#page=userstory/8794&searchPopup=userstory/8999)

This pr adds filtering to the search. It takes an optional `blocks` parameter formatted as an array of hashes e.g.

`[{identifier: 'client_group', value: 'Working age (18 - 65)'}] `

The search query is complex so it has been moved out of the controller into a class solely responsible for retrieving the documents.

I added brakeman to the gemfile to test that the sql query is secure. 

NEXT STEP
- [ ] Add validation on the number of filters that can be passed in the query

### DIFF
--- a/app/filters/document_provider.rb
+++ b/app/filters/document_provider.rb
@@ -1,12 +1,13 @@
 class DocumentProvider
-  attr_reader :current_site, :document_type, :keyword
+  attr_reader :current_site, :document_type, :keyword, :filters
 
   BLOCKS_TO_SEARCH = %w[content overview]
 
-  def initialize(current_site, document_type, keyword)
+  def initialize(current_site, document_type, keyword, filters)
     @current_site = current_site
     @document_type = document_type
     @keyword = keyword
+    @filters = filters
   end
 
   def retrieve
@@ -14,6 +15,7 @@ class DocumentProvider
 
     filter_by_type
     filter_by_keyword
+    filter_documents
   end
 
   private
@@ -22,8 +24,7 @@ class DocumentProvider
     return @documents if document_type.blank?
     
     @documents = @documents.joins(:layout)
-        .where('comfy_cms_layouts.identifier' => document_type)
-
+      .where('comfy_cms_layouts.identifier' => document_type)
   end
 
   def filter_by_keyword
@@ -33,5 +34,26 @@ class DocumentProvider
       .joins(:blocks)
       .where('comfy_cms_blocks.identifier' => BLOCKS_TO_SEARCH)
       .where('comfy_cms_pages.label LIKE ? OR comfy_cms_blocks.content LIKE ?', "%#{keyword}%", "%#{keyword}%").uniq
+  end
+
+  def filter_documents
+    return @documents if filters.blank?
+    
+    filters_to_hash.each do |filter, value|
+      @documents = @documents
+        .joins(:blocks)
+        .where('comfy_cms_blocks.identifier ? = AND comfy_cms_blocks.content = ?', filter, value)
+    end
+  end
+
+  def filters_to_hash
+    filters.reduce({}) do |acc, filter|
+      if acc[filter[:identifier]]
+        acc[filter[:identifier]] << acc[filter[:value]]
+      else
+        acc[filter[:identifier]] = acc[filter[:value]]
+      end
+      acc
+    end
   end
 end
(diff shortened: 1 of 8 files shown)


app/filters/document_provider.rb
```diff
@@ -0,0 +1,62 @@
+class DocumentProvider
+  attr_reader :current_site, :document_type, :keyword, :filters
+
+  BLOCKS_TO_SEARCH = %w[content overview]
+
+  def initialize(current_site, document_type, keyword, filters)
+    @current_site = current_site
+    @document_type = document_type
+    @keyword = keyword
+    @filters = filters
```
tomas-stefano: I think this PR is in a good state. All the tests now are clearer to understand each context and the data involved. My only question is related to the whole approach. Do you have a future plan on how the pagination will work? Will it be one more argument to this DocumentProvider or will be handled only on the controller?

tomas-stefano: By the way I know that the pagination will not be addressed on this PR, but would be good to start to think about it since it affects the current solution.

app/filters/document_provider.rb
```diff
@@ -0,0 +1,65 @@
...
+      .where('comfy_cms_layouts.identifier' => document_type)
+  end
+
+  def filter_by_keyword
+    return @documents if keyword.blank?
+
+    @documents = @documents
+      .joins(:blocks)
+      .where(
+        'comfy_cms_pages.label LIKE ? OR (comfy_cms_blocks.content LIKE ? AND comfy_cms_blocks.identifier IN (?))', "%#{keyword}%", "%#{keyword}%", BLOCKS_TO_SEARCH
```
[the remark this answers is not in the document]
margOnline: ♻️

spec/filters/document_provider_spec.rb
```diff
@@ -0,0 +1,234 @@
...
+    context 'when the search term is a phrase' do
+      let!(:page_with_phrase) { create(:insight_page, site: site, layout: insight_layout) }
+      let!(:page_without_phrase) { create(:page, site: site, layout: insight_layout) }
+
+      let(:keyword) { 'Financial well being: the employee view' }
+
+      it 'returns an array of documents which contain the phrase' do
+        expect(subject.size).to eq(1)
+        expect(subject).to match_array([page_with_phrase])
+
```
willhall88: delete line

spec/filters/document_provider_spec.rb
```diff
@@ -0,0 +1,234 @@
...
+            expect(subject).to match_array([page_with_keyword])
+          end
+        end
+
+         context 'and the keyword is found in a "content" block' do
+          let!(:page_with_keyword) { create(:insight_page_with_overview_block, site: site, layout: insight_layout) }
+          let!(:page_without_keyword) { create(:page, site: site, layout: insight_layout) }
+          let(:keyword) { 'redundancy' }
+
+          it 'returns documents which contain the keyword in an content block' do
```
willhall88: in `a` content block. 

I'd also consider rewording this and/or the context block? wdyt? same for the overview block content.
`and the keyword is found in a "content" block`
`returns documents which contain the keyword in an content block`

not a blocker / don't have to change the wording, it just seems long winded.

spec/factories/pages.rb
```diff
@@ -40,57 +40,62 @@
...
       label 'Debt, stress and pay levels'
       after(:create) do |page|
-        create :block, :debt_content, identifier: 'content', blockable: page
+        create :block, :debt_content, blockable: page
+        create :block, :working_age_client_group, blockable: page
       end
     end
 
-    factory :insight_page_about_pensions do
+    factory :uk_study_abt_work_and_stress do
```
willhall88: why have you abbreviated `about` ?

spec/filters/document_provider_spec.rb
margOnline: @willhall88 If the `describe`, `context` and `it` strings are read as a sentence then it would make sense to change the `it` string to something like `returns only those documents`, wdyt? otherwise what do you suggest?

willhall88: yeah, that would be how I would read it, and how it will appear when running rspec in documentation format. I'd be happy with that, but not a blocker if you think it should be kept as it is

willhall88:
:+1:

PR #415, opened 2018-03-21
TP 9090 Only allow a maximum of 26 filters as part of the API request

margOnline:
[TP 9090 ](https://moneyadviceservice.tpondemand.com/restui/board.aspx?#page=board/5580782241563718174&appConfig=eyJhY2lkIjoiRTQ4MDEyNENDOUYxRDU4RDg0RTEyRjNCODBEN0VBMUYifQ==&searchPopup=userstory/9090)

It is technically possible to provide an unlimited number of filters which would affect the database query since each filter generates a subquery. This pr limits the number of subqueries generated to 26. Refer to the TP card as to why its limited to 26.

### DIFF
--- a/app/controllers/api/documents_controller.rb
+++ b/app/controllers/api/documents_controller.rb
@@ -15,7 +15,11 @@ module API
         params[:blocks]
       ).retrieve
 
-      render json: documents, meta: { results: documents.size }, root: 'documents'
+      if documents
+        render json: documents, meta: { results: documents.size }, root: 'documents'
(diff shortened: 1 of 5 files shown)


margOnline:
@tomas-stefano I'll double check the requirements. Re testing, originally I was thinking a feature test in fincap would be the way to test this but have since decided maybe a request spec here is the way to go.

margOnline:
@tomas-stefano ♻️

PR #416, opened 2018-03-26
Index categories and pages for site search

tomas-stefano:
## Context
[TP card](https://moneyadviceservice.tpondemand.com/entity/8785-site-search-switchover)

This pr replaces the Google site search because Google will not support site search without the ads version. Algolia was chosen because ?

This is a time sensitive card and apologies for the big PR. I wish I have more time to break down into smaller PRs.

## Solution

The Adapter pattern was employed to provide flexibility, specifically to allow the business to change the indexer service if required.

For example there are two adapters:

* Local (prints the objects to be indexed - a very basic implementation).
* Algolia (send all the objects to be indexed to the Algolia service).

In order to create more adapters you need to create an object
that responds_to `#index_name` and `#create_or_update`.  You also need to set up
an environment variable `INDEXERS_ADAPTER` for your new adapter.

## Indexing categories

The current CMS database only has **one** record on the categories table that represents both the english and the welsh. An example:

```
#<Comfy::Cms::Category id: 129, site_id: 1, label: "about-us",
 title_en: "About us", title_cy: "Amdanom ni", 
description_en: "About us", description_cy: "Amdanom ni">
```

When indexing categories we need to define both an 'english' and a 'welsh' object for the category to be indexed.

## Indexing pages

Discussions are ongoing as to how to index the content of the pages. It is possible that as the discussion progresses, amendments will be required.

Because the current Algolia basic account does not support long content, it has been decided to use the meta description field as the content of the page to be indexed.

## Task

The index task will run every morning at 5 a.m.

There is also a rake task which you can run using the local adapter to see the objects that are generated by the indexers:

to Run: 

`$ INDEXERS_ADAPTER=local rake search:index`

This will generate categories and pages a table that looks like these:
```
+---------------+-------------------------------------------------------+
|              Index: categories - 192 number of indices                |
+---------------+-------------------------------------------------------+
| Field         | Value                                                 |
+---------------+-------------------------------------------------------+
| objectID      | /en/categories/about-us                               |
| title         | About us                                              |
| description   | About us                                              |
| links         | []                                                    |
+---------------+-------------------------------------------------------+
```

## Caveats

I really really try to add feature tests for the indexers but I fail miserably. [You can see the attempt here](https://github.com/moneyadviceservice/cms/tree/tests/add-feature-tests-for-indexers). I added the reasons of the fail attempt on the [commit message](https://github.com/moneyadviceservice/cms/commit/e65abde60f52c748b85d182a0338566aeaa9fdfe).

## Next Pull requests

- [ ] Replace site search on MAS site (frontend repo).
- [x] Create a PR on scripts repo which adds the env vars for Algolia for all environments.


spec/support/shared_examples/an_index_shared_example.rb
```diff
@@ -0,0 +1,32 @@
+RSpec.shared_examples 'an indexer' do |parameter|
```
margOnline: nice use of shared behaviour

spec/lib/cms/index_database_task_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+    it 'index selected pages' do
+      expect(page_indexer).to receive(:new).with(
+        collection: pages,
+        adapter: adapter
+      ).and_return(page_indexer)
+      expect(page_indexer).to receive(:index)
+      task.index_pages
+    end
+  end
+end
```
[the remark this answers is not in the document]
tomas-stefano: There are tests for the excluded pages here: https://github.com/moneyadviceservice/cms/pull/416/files#diff-912594a558faaf9351f9fdbeada9a828R39

tomas-stefano:
Hi @margOnline thank you for your comments. I really tried but it didn't work. There is more info on the PR description if you want to know more details. I will see until Thursday if there will be time to add a feature test for it.

lib/cms/indexers/category.rb
```diff
@@ -0,0 +1,23 @@
+module Indexers
+  class Category < Base
+		LOCALES = %w[en cy]
```
willhall88: indentation?

spec/lib/cms/indexers/page_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+  let(:adapter) { 'local' }
+
+  describe '#objects' do
+    let(:collection) do
+      [
+        create(:page,
+               label: 'Financial well being',
+               slug: 'financial-well-being',
+               meta_description: 'description',
+               published_at: Date.today,
```
willhall88: not a blocker, but FYI - using the latest rubocop *and* the yaml config from mas-standards rubocop would raise the following offense
```
page_spec.rb:17:35: C: Rails/Date: Do not use Date.today without zone. Use Time.zone.today instead.
```

willhall88:
:+1:

PR #423, opened 2018-04-16
[DEV APPROVED] Add feature tests for indexing cms database

tomas-stefano:
[TP card](https://moneyadviceservice.tpondemand.com/entity/9095-cms-add-feature-tests-for-indexing)

## Context

We are missing feature tests to check the proper indexing of our pages from our CMS system.  
 
This PR provides feature tests to make sure we are sending the right objects to be indexed on the site search.

## Technical Approach

Since the indexing is performed by command line rake tasks, I added the [aruba gem](https://github.com/cucumber/aruba) which adds feature tests for command line apps. 
To make this approach work, I needed to force an environment variable option which I set to the `local` adapter because we don't want to index anything to Algolia on our feature tests.

### Caveat
Aruba clashes with Spring because aruba uses the tmp directory to run the CLI interface and Spring needs to be in the Rails root directory. The following error was generated:

```
spring was unable to find your config/application.rb file on /tmp/aruba.
```

**I deleted spring from the rake script to resolve this issue.**

### DIFF
--- a/features/index_database.rake
+++ b/features/index_database.rake
@@ -1,0 +1,56 @@
+Feature: Index Database for Site Search
+  As a user
+  I want to search through the site
+  So I can find the content I am looking for
+
+  Scenario: Index Categories
+    Given I have the category
+      | Field          | Value                            |
+      | label          | budgeting-and-managing-money     |
+      | title_en       | Budgeting and managing money     |
+      | title_cy       | Cyllidebu a rheoli arian         |
+      | description_en | Advice on running a bank account |
+      | description_cy | Cyngor ar redeg cyfrif banc      |
+    When I run `rake search:index`
+    Then the output should contain:
+    """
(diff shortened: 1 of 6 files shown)


margOnline:
Please could you update the description, its not very readable in its current form.

features/index_database.rake
```diff
@@ -0,0 +1,56 @@
+Feature: Index Database for Site Search
+  As an user
```
margOnline: `As a user`

one of the English language's many inconsistencies 😉

features/index_database.rake
```diff
@@ -0,0 +1,56 @@
...
+  I want search through the site
```
margOnline: `I want to`

tomas-stefano:
@margOnline I updated with more explanation. is it better now?

PR #427, opened 2018-04-18
[WIP] Update DB yaml

osharpe:
Update yaml file in Fincap Branch


config/database.yml
```diff
@@ -27,7 +27,7 @@ production:
   encoding: utf8
   reconnect: true
   strict: false
-  url: <%= ENV["MAS_CMS_DATABASE_URL"] %>
+  url: <%= ENV["MAS_FINCAP_CMS_DATABASE_URL"] %>
```
tomas-stefano: @osharpe how the cms and fincap will be sort here? If we merge on fincap branch at some time this will be merged on master and affect MAS CMS as well. Wdyt? And @pmcostello thinks as well?

osharpe: Yes this is correct and should not be merged to the master branch yet. Not sure how we will manage this in production. We could simply have two separate yamls, which get renamed during the build but I don't really like this as a solution.

PR #428, opened 2018-04-18
[DEV APPROVED] Feature/9153 cms article creation form

tomas-stefano:
[TP card](https://moneyadviceservice.tpondemand.com/entity/9153-cms-article-creation-form)

## Context

This PR addressed the card of creating an Fincap article on CMS including the components.

There are currently 4 different types of component, markup examples and UI design:

* CTA Component (call to action links)
* Downloads Component (List of links to be downloaded)
* Feedback Component (An email to receive feedback)
* Hero section (Including the image and the description)

Each component form requires a textarea with an example placeholder for the expected values (to aid content editors in the creation of the fincap article pages).

## Solution
I defined a new Comfy gem `tag` which I've named `simple component`. 

How the Comfy gem works: given you have a Layout object with content `'{{ cms:page:hero_image:simple_component/example  }}'` the Comfy gem will collect all classes under the ComfortableMexicanSofa::Tag namespace, scan the content and check the comfy tags that match this regular expression. Every time you add a ComfortableMexicanSofa::Tag (like the simple component) the class will be available on the layout objects.

This component will accept any content and the client will decide how to render it.
The reason for the name "simple_" is because the client will decide how to render the components and not CMS itself, which makes the implementation here quite easy.


features/step_definitions/fincap_pages_steps.rb
```diff
@@ -26,6 +26,30 @@
   )
 end
 
+Given(/^I have an article page layout setup with components$/) do
+  english_site = Comfy::Cms::Site.find_or_create_by(
```
willhall88: the set up of the site is already carried out at the top step `Given(/^I have an insight page layout$/) do`
Perhaps we can split that into it's own step, that we only call from the steps themselves? wdyt?

tomas-stefano: Oh, I forgot about that line. Yes good idea.

tomas-stefano: @willhall88 actually I discover with @margOnline a helper called `cms_site` to be used on cucumber steps so I just used ... and it worked! 🎩 Wdyt? ♻️

willhall88: 😻

app/controllers/pages_controller.rb
```diff
@@ -72,13 +72,22 @@ def save_page
     # First change state if needed. We do a non saving event so we can access
     # the dirty logging in the content registers.
     @page.update_state(params[:state_event]) if params[:state_event]
+    blocks_attributes = params[:blocks_attributes]
```
margOnline: 😍

PR #430, opened 2018-04-26
[DEV APPROVED] Set search:index rake task to every day 5am

osharpe:
Update cron schedule to run Algolia index rake task everyday at 5am.

### DIFF
--- a/config/schedule.rb
+++ b/config/schedule.rb
@@ -19,6 +19,6 @@
 
 # Learn more: http://github.com/javan/whenever
 
-every 4.days, at: '5:00am' do
+every 1.day, at: '5:00am' do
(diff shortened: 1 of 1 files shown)


willhall88:
are we increasing this back to every 1 day as we have now solved the issue of running into our monthly limit with Algolia?

willhall88:
:+1: just the above question. code is fine

osharpe:
@willhall88 Ye we were previously using `clockwork` gem which ran the rake task on both `cms01` and `cms02`, effectively doubling the amount of indexing operations required. We have now moved to `whenever` gem which is specified just to run on just `cms01` in scripts repo. We are coming close to billing cycle, so we will have a fresh quota, it's now safe to move back to the daily cycle.

PR #431, opened 2018-04-27
[DEV APPROVED] Add meta keywords field to site search index

tomas-stefano:
[TP card](https://moneyadviceservice.tpondemand.com/entity/9144-site-search-enhancement-add-meta-keywords)

## Context

The Site Search product, Algolia, is working on the site but we are limited in the tools we need in order to influence search results. 
Currently Algolia indexes pages on page title and meta-description.
We could edit meta descriptions in order to help fix searches with missing results, and to help influence what appears towards the top of the results. However, the problem with this is that Meta description is used on Google SERPs, and generally full sentences are better than keyword stuffing for these.

So adding a field to add those keywords makes the site search flexible for the Digital team to influence the search results.

## Solution

So with the solution, I added a field called meta keywords and the way that will work on the site search is creating sentences separated by comma.

So for example Imagine that I have the following on the meta keywords:

site search

And on Algolia I have the following configuration (keywords the first attribute to be searchable):

algolia

When I search for "Budget Planner" the keywords that I added has more relevance so it appears first:

screen shot 2018-04-27 at 17 10 21

The same with budgeting:

screen shot 2018-04-27 at 17 10 37

### DIFF
--- a/app/views/pages/_form.html.haml
+++ b/app/views/pages/_form.html.haml
@@ -32,6 +32,13 @@
           = render 'tags/shared/form',                         form: form
           = render 'form_meta_description',                    form: form
           = render 'form_meta_title',                          form: form
+          .form-group
+            .form__label-heading
+              = form.label :meta_keywords
+            .form__input-wrapper
+              = form.text_field :meta_keywords, class: 'form__input form__input--has-icon'
+              %span.form__hint
+                = Comfy::Cms::Page.human_attribute_name(:meta_keywords_description)
           = render 'form_additional_options',                  form: form
 
     = render partial: 'form_blocks'
--- a/config/locales/pages.yml
+++ b/config/locales/pages.yml
@@ -2,6 +2,8 @@ en:
   activerecord:
     attributes:
       comfy/cms/page:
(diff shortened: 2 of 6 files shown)


tomas-stefano:
1. Fair point, but the idea came from the editors themselves to use a comma. Sorry I should have put on the PR description about that.
2.The comfy gem takes care by permitting any parameters from `:page`: https://github.com/moneyadviceservice/comfortable-mexican-sofa/blob/master/app/controllers/comfy/admin/cms/pages_controller.rb#L144

margOnline:
@tomas-stefano thx for the clarification. re point 1, I still think some ui guidance should be added. What about adding some text with an example as per the the url form field. Perhaps a question for the product manager?

tomas-stefano:
@margOnline I aligned with the editors and Ed and they are requested the wording. I changed and pushed. Can you review, please?

willhall88:
Looks like there has been plenty of discussion with the editors anyway, but I'm assuming it's been made clear these keywords are for our Algolia site search *only* and *not* for Googles SERP?

PR #446, opened 2018-05-23
9205 - Thematic review

tomas-stefano:
[TP card](https://moneyadviceservice.tpondemand.com/entity/9205-thematic-reviews-content-association)

## Context

This PR addresses the search through pages by tag.

Allowing the documents API  to search by tag means that on CMS we can search all summaries that are within a tag. The tag will be associated with a Thematic review because the name of the tag and the name of the slug are the same.

## Example

Given that I have the thematic review:
```
Type: Thematic Review
Slug: my-awesome-thematic-review-page
```

and an Insight:

```
Type: Insight
slug: some-insight
tag: my-awesome-thematic-review-page
```

Note that the Insight above  **should** have a tag with the same name as the thematic review's slug.

When we search for the tag "my-awesome-thematic-review-page", the pages which have this tag will be returned. On this example,  the "some-insight" page will be included.

## Technical

This PR includes real data to the components of the thematic review to facilitate navigating to the hub with the real tag.

#### Refactoring
I refactored the DocumentProvider names. The terms `page type`  and `document type` were being used interchangeably,  making debugging more difficult. I think now it is more concise.

### DIFF
--- a/app/controllers/api/documents_controller.rb
+++ b/app/controllers/api/documents_controller.rb
@@ -8,9 +8,10 @@ module API
 
     api :GET, '/:locale/documents'
     param :locale, String, required: true
-    param :page_type, String, required: false
+    param :document_type, Array, required: false
     param :keyword, String, required: false
     param :blocks, Array, required: false
+    param :tag, String, required: false
     def index
       if documents
         render json: paginated_documents, meta: meta_data, root: 'documents'
@@ -23,10 +24,12 @@ module API
 
     def documents
       @documents ||= DocumentProvider.new(
-        current_site,
-        params[:document_type],
-        params[:keyword],
-        params[:blocks]
+        params.permit(
+          :keyword,
+          :tag,
+          document_type: [],
+          blocks: [:identifier, :value]
+        ).merge(current_site: current_site)
       ).retrieve
     end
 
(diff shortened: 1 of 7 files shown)


spec/controllers/api/documents_controller_spec.rb
```diff
@@ -75,13 +75,81 @@
...
+        )
+      end
+      let!(:page_without_tag) do
+        create(:page, site: site)
+      end
+      let!(:page_with_a_different_tag) do
+        create(:page_with_tag, site: site, tag_name: 'sorry-not-a-pensions')
+      end
+
+      context 'when searching by a tag' do
```
margOnline: the nested context description is the same as its parent, please amend one or the other to make more readable

tomas-stefano: ♻️

app/controllers/api/documents_controller.rb
```diff
@@ -23,10 +24,12 @@ def index
...
       @documents ||= DocumentProvider.new(
-        current_site,
-        params[:document_type],
-        params[:keyword],
-        params[:blocks]
+        params.permit(
+          :keyword,
+          :tag,
+          document_type: [],
+          blocks: [:identifier, :value]
```
willhall88: 🙌

app/filters/document_provider.rb
```diff
@@ -1,14 +1,15 @@
...
 
   BLOCKS_TO_SEARCH = %w(content overview)
   FILTER_LIMIT = 26
 
-  def initialize(current_site, document_type, keyword, filters)
-    @current_site = current_site
-    @document_type = document_type
-    @keyword = keyword
-    @filters = filters
+  def initialize(params = {})
```
willhall88: 🙌

PR #471, opened 2018-08-03
[DEV APPROVED] Fix for  teaser_section_title not displaying on Fincap

margOnline:
The title for the teaser section with the 3 teaser boxes is not displaying. 

### TECHNICAL
There are 2 solutions to fixing the issue of the teaser_section_title
not displaying:

1. change the field to be a simple text field and not a component
2. update the identifier that the article_template is looking for in fincap

I chose the first approach because it made more sense that this field be a simple text field since the content will only ever be a few words of text.


tomas-stefano:
The code lgtm but why it should be a text field? Can you elaborate and add this info to the commit message please?

margOnline:
@tomas-stefano ♻️

PR #478, opened 2018-08-16
[DEV APPROVED] 9245 UK strategy page layout

your:
[TP 9245](https://moneyadviceservice.tpondemand.com/entity/9245-cms-uk-strategy-page)

One of the pages required in the CMS by FinCap is the "UK Strategy" page.  This exists in the current live site (but is very simple). This PR creates a new page with a new layout, additional content and navigation.

This PR is required for [TP 9246](https://moneyadviceservice.tpondemand.com/entity/9246-uk-strategy-page-view).


db/seeds/fincap_data/article_seeds.rb
```diff
@@ -20,108 +20,6 @@
...
-      [Financial Capability Progress Report 2017](/fincap+progress+report+2017.pdf)
-      CONTENT
-    ),
-    Comfy::Cms::Block.new(
-      identifier: 'component_feedback',
-      content: 'email@moneyadviceservice.org.uk'
-    )
-  ]
-)
-
```
tomas-stefano: Will we not have an article page created on seeds then?

your: @tomas-stefano that's a good point, maybe we can seed another article, but definitely with some other content (unrelated to UK Strategy). What do you think?

margOnline: Good point, maybe create the 'our-approach' article as it is required for one of the [homepage ctas](https://github.com/moneyadviceservice/fin_cap/blob/master/app/views/homepages/show.html.erb#L47).

tomas-stefano: Is there any tests on fincap that relies on this data?

margOnline: If you mean 'our-approach' article, I think not

your: I'm adding it then, it sound sensible to have one anyway. 👍

margOnline: @your I think it should be added as a separate pr, strictly speaking it should have been part of [this pr](https://github.com/moneyadviceservice/fin_cap/pull/174).

your: @margOnline Fair enough, it would require [this test on fincap](https://github.com/moneyadviceservice/fin_cap/blob/master/features/article_page.feature) to be adapted anyway (after removing the cassette), so I would agree with you. Can do in a separate ticket/PR.

margOnline: @tomas-stefano other than the test for `our-approach` which I'm going to pick up, can this pr be approved?

margOnline: @your  I just realised that the fincap:setup task also needs to be amended to include the creation of the uk_strategy layout, refer to [pr #465](https://github.com/moneyadviceservice/cms/pull/465/files#diff-fc7b15bcd327a3dc679c3acba3682f47)

your: @margOnline thanks for spotting it, added with d05a94b

PR #480, opened 2018-08-22
[DEV APPROVED] Add seeds for redirects on fincap app

tomas-stefano:
[TP card](https://moneyadviceservice.tpondemand.com/entity/9225-implement-internal-redirect-feature-on-fincap)

This PR adds seeds for the PR https://github.com/moneyadviceservice/fin_cap/pull/184

### DIFF
--- a/db/seeds/fincap_data/redirects.rb
+++ b/db/seeds/fincap_data/redirects.rb
@@ -1,0 +1,24 @@
+[
+  {
+    source: '/document/WOuEzCsAAIaQ8tXy/financial-well-being-the-employee-view',
+    destination: '/en/insights/financial-well-being-the-employee-view'
+  },
+  {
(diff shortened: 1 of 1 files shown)


db/seeds/fincap_data/redirects.rb
```diff
@@ -0,0 +1,24 @@
...
+  {
+    source: '/young-adults',
+    destination: '/en/lifestages/young-adults',
+  },
+  {
+    source: '/en/news/old-news',
+    destination: '/en/news/press-release-a-new-way-to-pay'
+  }
+].each do |attributes|
+  Redirect.create!(attributes.merge(redirect_type: 'permanent'))
```
your: I don't think we should bang this, seeds should be run idempotently, as many times as you want, right?

tomas-stefano: Yes, that's true. I updated using `find_or_create-by`. ♻️

PR #482, opened 2018-08-24
[DEV APPROVED] - Make redirect works for MAS and FINCAP app

tomas-stefano:
[TP card](https://moneyadviceservice.tpondemand.com/entity/9225-implement-internal-redirect-feature-on-fincap)

## Problem

Fincap uat was redirecting to the MAS site. One example you can see https://fincap-cms.uat.dev.mas.local/api/en/evaluations/evaluation-of-my-money-now

## What should happen

Mas CMS redirects should redirect to MAS site.
Fincap CMS redirects should redirect to Fincap site.

## Investigation

There was an env var hard coded on redirect middleware which uses as the domain to be redirect to.

Make the domain to look for the right APP_NAME makes the redirect work as expected to.

## Considerations

The env var that's being used  in MAS site is called FARADAY_HOST. I think is a terrible name but I decided to leave as it is to not break functionality under the MAS site.
There is an env var on production already called FINCAP_PUBLIC_WEBSITE_DOMAIN. I asked Oli and he confirmed that this is the env var that should be using.

### DIFF
--- a/config/initializers/domain.rb
+++ b/config/initializers/domain.rb
@@ -28,7 +28,8 @@ module Domain
                     :dev_db_name,
                     :prod_db_url,
                     :public_website_domain,
-                    :public_website_url
+                    :public_website_url,
+                    :redirect_domain
 
       def initialize
         @algolia_app_id = nil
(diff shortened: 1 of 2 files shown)


osharpe:
Good spot @tomas-stefano  👍

PR #483, opened 2018-08-24
[DEV APPROVED] [9508] Add seeds to allow feature testing on paginated search

MarcoBgn:
[TP card](https://moneyadviceservice.tpondemand.com/restui/board.aspx?#page=userstory/9508)

## Context

In order to enable feature tests for the Fincap Search feature, a new app has been created in Algolia:
`FINCAP_LOCAL_TESTING`

This allows to have a shared and common index based on the FINCAP Cms DB seed. 

In order to test pagination more records were needed. I have tried to keep it as simple as possible as this will be used mainly for testing

## To re-seed

Instructions to reset the seed have been added to the readme. 

## Considerations

Dropping the DB can be dangerous as you might have to re-create users, etc. The ideal solution, in the future, would be to use a `find_or_create` seeding mechanism

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ So the next section will explain the setup for each application.
 
 ### Environment setup (Mac OsX)
 
-MySql needs to be installed and linked. 
+MySql needs to be installed and linked.
 Make sure Homebrew is installed.
 
 ```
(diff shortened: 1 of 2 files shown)


your:
> Dropping the DB can be dangerous as you might have to re-create users, etc. The ideal solution, in the future, would be to use a find_or_create seeding mechanism

I agree on this, since seeds should be be run whenever you want without having to to drop the whole database first. Although, there are reasons for not using `find_or_create_by` and are mainly related to the fact that you need to destroy and recreated the pages if you change their layout.

PR #488, opened 2018-09-07
[DEV APPROVED] 9404 - Add a new What Works page layout

MarcoBgn:
[TP Card](https://moneyadviceservice.tpondemand.com/restui/board.aspx?#page=board/5682182231901749200&appConfig=eyJhY2lkIjoiNzEyRUZDMzlENEJERUE4QjJBRDVEQzdENUE1MEE2OEIifQ==&boardPopup=userstory/9404/silent)

### Changes
- Add a new layout (`What Works`). This is a copy of the `Lifestages` layout
  This will provide a new URL scope. i.e /en/what_works/my-custom-page
- Amend fincap:setup rake task
- Add `what_works` seed data 

### Fix/Refactor

- We have discovered that tags were not properly added in the seeds. We have also simplified the logic, avoiding a second iteration over the newly created pages.

### DIFF
--- a/db/seeds/fincap_data/what_works_seeds.rb
+++ b/db/seeds/fincap_data/what_works_seeds.rb
@@ -1,0 +1,162 @@
+english_site = Comfy::Cms::Site.find_or_create_by(
+  label: 'en',
+  identifier: 'money-advice-service-en',
+  hostname: ENV['HOSTNAME'] || 'localhost:3000',
+  path: 'en',
+  locale: 'en',
+  is_mirrored: true
+)
+
+what_works_layout = english_site.layouts.find_or_create_by(
+  identifier: 'what_works',
+  label: 'What Works',
+  content:  <<-CONTENT
+    {{ cms:page:content:rich_text }}
+    {{ cms:page:hero_image:simple_component/https://moneyadviceservice.org.uk/image.jpg }}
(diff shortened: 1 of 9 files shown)


MarcoBgn:
Good point @tomas-stefano, What Works seed added

scruti:
Added 7f46504 fixing an inflection issue with "what works" when called from CMS Client. Will squash it into the feature commit before merging it.

db/seeds/fincap_data/what_works_seeds.rb
```diff
@@ -0,0 +1,162 @@
+english_site = Comfy::Cms::Site.find_or_create_by(
+  label: 'en',
+  identifier: 'money-advice-service-en',
+  hostname: ENV['HOSTNAME'] || 'localhost:3000',
```
your: Question unrelated to this PR probably: I cannot find this `HOSTNAME` anywhere in the [.env-example](https://github.com/moneyadviceservice/cms/blob/master/.env-example) – I see it's present in other seed files, but is it actually is use?

MarcoBgn: Good point, I believe it is set depending on the env:
```bash
[root@***-tst-**** ~]$ echo $HOSTNAME
***-tst-****.dev.mas.local
```

PR #489, opened 2018-09-11
[DEV APPROVED] Extract all Fincap Layouts into a config file

scruti:
During the setup of a new Fincap layout we realised that we were
pasting the same block of content in three different places:
- Seeds File
- Fincap rake task
- Feature test

Having to paste the same block of content in those places every time
we add a new layout seems to me annoying and prone to errors.
Due to that, this is extracting the new layouts setup into a single
config file, and the same configuration can be used later in multiple
places without having to paste the same content around.


lib/tasks/fincap.rake
```diff
@@ -14,233 +14,11 @@ namespace :fincap do
...
-          {{ cms:page:teaser3_image }}
-          {{ cms:page:teaser3_text }}
-          {{ cms:page:teaser3_link }}
-          {{ cms:page:horizontal_teaser_title }}
-          {{ cms:page:horizontal_teaser_image }}
-          {{ cms:page:horizontal_teaser_text }}
-          {{ cms:page:horizontal_teaser_link }}
-        CONTENT
-      )
+    layouts.each do |_name , attributes|
```
MarcoBgn: Nice one 👍

config/fincap_layouts.yml
```diff
@@ -0,0 +1,216 @@
...
+    {{ cms:page:teaser3_title }}
+    {{ cms:page:teaser3_image }}
+    {{ cms:page:teaser3_text }}
+    {{ cms:page:teaser3_link }}
+    {{ cms:page:strategy_title }}
+    {{ cms:page:strategy_overview }}
+    {{ cms:page:strategy_link }}
+    {{ cms:page:steering_group_title }}
+    {{ cms:page:steering_group_links }}
+    {{ cms:page:download:simple_component/[Text Link](https://moneyadviceservice.org.uk/link) }}
```
MarcoBgn: Is the missing newline on purpose?

scruti: nope. Good catch!

scruti:
This was dead for a while as I was checking it during some free time between tasks. 
The main blocker was some feature tests that were failing on this branch but not in master. 
Yesterday I managed to sort them out.

So... ready for review!

tomas-stefano:
Consider my message above a parrot_wave!

PR #491, opened 2018-09-13
[DEV APPROVED] - [9580] Fix multiple tags error and duplicated news

MarcoBgn:
- filter_by_tag was not allowing multiple tag.values using
  'value =' instead of 'value IN'.
- News were duplicated if multiple tags were present, so
  a 'uniq' => 'DISTINCT' clause has been added both in
  filter_by_tag and order_documents filters.

### DIFF
--- a/spec/factories/pages.rb
+++ b/spec/factories/pages.rb
@@ -30,6 +30,18 @@ FactoryGirl.define do
       end
     end
 
+    factory :page_with_multiple_tags do
+      site { create :site, identifier: 'test-documents' }
+      label 'A page with two tags'
+      after(:create) do |page, evaluator|
+        page.keywords.concat [
+          create(:tag, value: "#{evaluator.tag_name}-2"),
+          create(:tag, value: "#{evaluator.tag_name}-3")
+        ]
+        create :block, identifier: 'content', blockable: page
+      end
+    end
+
     factory :english_article do
       site { create :site, label: 'en' }
       layout { create :layout, identifier: 'article' }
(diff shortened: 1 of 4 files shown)


spec/filters/document_provider_spec.rb
```diff
@@ -277,7 +277,7 @@
         order_by_date: '2017-03-16'
       )
     end
-    let(:tag) { ['test'] }
+    let(:tag) { ['test', 'another-test'] }
```
your: Can we test the filtering with multiple tags in isolation? This context here is more about the ordering, so it tries to use a combination of filters (and I've actually added the tag here – only one 😞– for completeness sake...), but the main goal is to test the ordering indeed.

Something like "when filtering by tags" maybe? And/or maybe extending https://github.com/moneyadviceservice/cms/blob/e987fc050f22d6e733018e3cf2860ae6d36760a2/spec/controllers/api/documents_controller_spec.rb#L112-L141 as well?

MarcoBgn: Yes, I agree. Might not be clear why two tags have been added to the factory and in the future the context (or lack thereof 😄  ) might be lost 👍

MarcoBgn:
Thanks for the review and input @your , I have amended the commit

spec/factories/pages.rb
```diff
@@ -30,6 +30,18 @@
       end
     end
 
+    factory :page_with_multiple_tags do
+      site { create :site, identifier: 'test-documents' }
+      label 'A page with two tags'
+      after(:create) do |page, evaluator|
+        page.keywords.concat [
+          create(:tag, value: "#{evaluator.tag_name}-2"),
```
your: Nice!

tomas-stefano:
@MarcoBgn Nicely done! Don't forget to create a TP card and tell the Q.A/B.As ;)

PR #495, opened 2018-09-25
9594 fix evidence summaries created at

scruti:
[TP-9594](https://moneyadviceservice.tpondemand.com/entity/9594-evidence-summaries-created_at-updates)
### What does this PR
Adds a rake task that will find all the Evidence Summaries pages published prior to 2018 and replace their `created_at` to the first second of their year of publication.
Eg:  Year of publication: `2017` --> `created_at`: `2017-01-01 00:00:0`  

### Why do we need this change?
After implementing the filtering options in the new [Fincap Evidence Hub](https://fincap.moneyadviceservice.org.uk/en/evidence_hub), was discovered that quite old pages were being returned as first results.

This was happening because the CMS is returning the pages ordered by their creation date (`created_at`), and that value was set to August 2018 for all the pages when they were migrated to the new Fincap site.

Due to that, old pages with years of publication like `2003`... `2010`... where being ordered on the results as `2018` content.

To avoid this, the decision was to override the `created_at` for all those evidence summary pages and set it up at the first second of their year of publication.

### Where should the reviewer start
For better context the tests of the rake tasks describe the expected behaviour.
The rake task itself is just a SQL Update query.

### Anything else to add? 
The PR includes a refactor commit over the Insight pages factories extracting some common behaviour and setting the correct layout to accommodate further needed changes in the feature commit.


spec/lib/tasks/fincap_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+
+RSpec.describe 'fincap' do
+  before :all do
+    Rake.application.rake_require 'tasks/fincap'
+    Rake::Task.define_task(:environment)
+  end
+
+  describe 'fix_evidence_summaries_created_at' do
+    subject(:run_task) do
+      Rake::Task['fincap:fix_evidence_summaries_created_at'].reenable
```
MarcoBgn: Nice!

spec/lib/tasks/fincap_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+      let(:year_of_publication) { nil }
+      include_examples 'not updated by the rake task'
+    end
+
+    context 'when the year of publication is empty' do
+      let(:year_of_publication) { '' }
+      include_examples 'not updated by the rake task'
+    end
+  end
+end
```
margOnline: Good set of test cases

scruti:
Thanks for the input, and is a really good point.

On this subject my approach is: 
- If this is just this case/not recurring logic and we need this concrete db timestamps update, I just create it, execute it and done.
- If we get new requirements with some variations like further executions with different years... time to extract an abstraction that allows this sensibility.

MarcoBgn:
Good points @margOnline , I agree generally on all the above. 

In this instance though, considering this being a fairly specific case, I think the cost of extracting the task (which is basically an SQL statement) might outweigh the benefits

margOnline:
Thx all for the comments, it appears we have the same general opinion. Is the intention to delete this task once it’s been run as @your indicated?

scruti:
I'm personally keen to delete it because I don't see this as a recurring operation (and in case of needing a similar one we have git history for recovering it) and, specially, we have permanent tests hitting DB ans slowing down the full suite for one/off stuff that probably won't have further executions.

I'm happy to revert the Feature commit once this is executed and the issue is fixed.

What do you think?

scruti:
Added a new commit (5149d4e) as discovered when testing this in UAT that the `created_at` change wasn't actually affecting to the order of returned Evidence Summaries as the order was by "position" page attribute and then "created_at".

I'm not going to call it "a bug" as I assume the "position" was automatically  increased on the new pages and the fact that we actually manually modified the `created_at` kinda of messed it up. Anyways covered it by a test to show the point that before the change the position was causing a red test and after the `reorder` change we got a green test.

Happy to get some views/proposals/alternatives on this change as I have almost 0 context on this default behaviour and are kind of working on assumptions based on my queries/tests.

PR #498, opened 2018-10-15
[DEV APPROVED] [9418] [9421] autogenerate rubocop config cms

MarcoBgn:
[TP Card](https://moneyadviceservice.tpondemand.com/restui/board.aspx?#page=board/5682182231901749200&appConfig=eyJhY2lkIjoiNzEyRUZDMzlENEJERUE4QjJBRDVEQzdENUE1MEE2OEIifQ==&boardPopup=task/9418/silent)

# Context
- Bump to the latest minor version, conservative
- Add cascading rules from 'mas-standards' > 'cms-specific rules' > `.rubocop_todo.yml` 
- Add hooks and README instructions to run tests before pushing
- Apply `--auto-correct` to the `Layout` namespace cops 

## Considerations
- New violations will now be picked up
- The `--auto-correct` flag seems to be messing quite heavily with the existing specs. I think will be good to add separate PRs addressing the violations by Blocks/Namespace or this will become very big. What do you think?


scruti:
This changes are great, thanks for jumping into it! :clap: 

> The --auto-correct flag seems to be messing quite heavily with the existing specs. I think will be good to add separate PRs addressing the violations by Blocks/Namespace or this will become very big. What do you think?

I would definitely break them into separate work. The update && TODO has enough value to be delivered by itself. Cleaning up TODO stuff can be done progressively in the future.

PR #503, opened 2019-01-14
[DEV APPROVED] 10162 - Add audit task

tomas-stefano:
[Tp Card](https://moneyadviceservice.tpondemand.com/entity/10162-pull-list-of-cms-content-to)

## Context

- CSV file created:  `/tmp/audit.csv`
- All fields required are populated, example `csv` added in the TP card

The task is called audit:pages and it can be run locally as: `rake audit:pages` which will generate a csv 

## Testing strategy

- A bit of setup is required due to the CMS domain model 
  
  - A `layout` has to be present in the page factory to provide an `identifier` ('video', 'article', 'corporate')
  - A `Comfy::Cms::Site` has to be present in order to retrieve the language 
  - A `Comfy::Cms::Categorization`'s table has to be present in order to retrieve the `category_names`
  - A `Comfy::Cms::Revision` has to be present in order to retrieve the `author` name (`Comfy::Cms::User`) through `ActivityLog`
 - a `Tempfile` has been used to avoid needlessly adding files to the server

**note** Language has been added as a required field 

You can run on the test environment as:

```
ssh your.user@sometestenv.dev.mas.local
sudo su -
cd /srv/cms
. /etc/mas/environment
rake audit:pages
mv /tmp/audit.csv /home/your.home.directory
exit
exit
# On your local machine
scp your.user@sometestenv.dev.mas.local:~/audit.csv .
```
And then you have the csv with you.

Tasks that needs to be done.

- [ ] Needs to deploy to a test environment.
- [ ] Run the task there and check if the CSV is generated and contain the info that the card asks.
- [ ] Ask a Q.A to check the csv file.
- [ ] Deploy to staging/production.


spec/factories/revisions.rb
```diff
@@ -13,4 +13,19 @@
       }
     }
   end
+
+  factory :revision_with_event, class: Comfy::Cms::Revision do
+    transient do
```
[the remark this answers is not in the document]
MarcoBgn: It is used to pass in parameters, on initialisation,  that are not 'real' fields. The `data` field is a hash:
`data: {"content"=>nil, "css"=>nil, "js"=>nil, "author"=>{}}`. 

Having a transient block allows to create a `Revision` object with fields associated with a `User`:
```
create(
      :revision_with_event,
      record: page_with_meta,
      user_id: current_user.id,
      user_name: current_user.name
    )
```

MarcoBgn:
Thanks for the quick review @margOnline. Sure, I will try to make it clearer

MarcoBgn:
@margOnline comments addressed. I have added some more info in the PR description and in the commits.

spec/lib/cms/audit/pages_spec.rb
```diff
@@ -0,0 +1,111 @@
+describe Cms::Audit::Pages do
+  let(:role) { :user }
+  let(:current_user) { create(:user, role: Comfy::Cms::User.roles[role]) }
+  let(:state) { 'published' }
+  let(:en_site) { create(:site) }
+  let(:cy_site) { create(:site, label: 'cy') }
+  let(:page_with_tags) { create(:page_with_multiple_tags, site: en_site, state: state) }
+  let(:page_with_meta) { create(:article_with_metadata, site: cy_site, state: state) }
+  let!(:revision) do
```
scruti: For all this `let!` declarations (`:revision`, `revision_2`, `category` and `categorization`) as it seem to be context setup that is not reused anywhere (as far as I've seen they have no calls)... have you thought about the possibility of avoiding the boilerplate of the `let` and move them to a block where `create this in DB before executing the tests` falls as a natural thing?
```
before do
  Timecop.freeze(Time.zone.local(1990))
  create(
      :revision_with_event,
      record: page_with_tags,
      user_id: current_user.id,
      user_name: current_user.name
    )
  create(
    :revision_with_event,
    record: page_with_meta,
    user_id: current_user.id,
    user_name: current_user.name
  )
  create(
    :categorization,
    categorized_id: page_with_tags.id,
    category: create(:category, site: en_site),
    categorized_type: 'Comfy::Cms::Page'
  )
end
```
 In this way we can see explicitly what are the main thing that need to be set up in order to execute the test. While the `let` declarations outside the block are helpers/accessories to the context setup.

What do you think?

spec/lib/cms/audit/pages_spec.rb
```diff
@@ -0,0 +1,111 @@
...
+  describe 'self.generate' do
+    subject { described_class.generate_report(file: temp_file) }
+
+    let(:temp_file) { Tempfile.new('temp_file.csv') }
+
+    it 'generates a csv file and returns the count of pages added' do
+      expect(subject).to eq 2
+    end
+
+    context 'csv file content' do
```
scruti: nitpick: It looks like this, being a specialisation of what to check and not a "when this happens.../with this setup...", suits better a `describe` block than a `context` (even if they're actually the same)

spec/lib/cms/audit/pages_spec.rb
MarcoBgn: Yep, good point! Pushing up in a minute

spec/lib/cms/audit/pages_spec.rb
```diff
@@ -0,0 +1,111 @@
...
+          'false',
+          'true',
+          '01/01/1990, 00:00',
+          current_user.name,
+          state,
+          'cy'
+        ]
+      end
+
+      it 'creates a file with correct headers' do
```
scruti: Totally subjective but in these tests I personally would not rely on a `before` block to execute the `subject under test` and posterior CSV parsing and asserted fields, but go for explicitness  calling both  `audit_generation` (or whatever, I would go for a named subject there), CSV parsing and then the expected fields all together inside each test so I don't need to scroll around to check what is declared for the tests when reading them, especially when `expected_page_`\2` are not reused anywhere so don't need to be extracted from individual tests.

MarcoBgn: I tried it but to be honest it looks messier to me 😄 Especially the array after the `expect`. Unless someone has a very strong opinion against it I would rather leave the variables

MarcoBgn:
Thanks for the inputs @margOnline @scruti. I have addressed the comments ♻️

PR #504, opened 2019-01-21
[DEV APPROVED] Use UTF-8 encoding to avoid incompatibilities

MarcoBgn:
Related to: [TP Card](https://moneyadviceservice.tpondemand.com/restui/board.aspx#page=board/5682182231901749200&appConfig=eyJhY2lkIjoiRkRBMzk3RTBGNjU3NDg1RDM4QzczMTJFNzI5MDFBN0MifQ==&boardPopup=userstory/10162/silent)

- After deployment I have tested the task and discovered
  incompatibilities with the encoding. Using UTF-8 fixes the issue

- Generated file attached to the TP Card

### DIFF
--- a/spec/lib/cms/audit/pages_spec.rb
+++ b/spec/lib/cms/audit/pages_spec.rb
@@ -87,6 +87,10 @@ describe Cms::Audit::Pages do
         ]
       end
 
+      it 'used UTF-8 encoding' do
+        expect(described_class::ENCODING).to eq 'UTF-8'
+      end
+
       it 'creates a file with correct headers' do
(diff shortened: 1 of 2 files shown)


spec/lib/cms/audit/pages_spec.rb
```diff
@@ -87,6 +87,10 @@
         ]
       end
 
+      it 'used UTF-8 encoding' do
+        expect(described_class::ENCODING).to eq 'UTF-8'
+      end
```
scruti: This is one of the few times you will see me saying "you don't need a test here" :laughing: 
Basically we're testing a constant value, not business logic, so becomes a little bit redundant IMHO. 
But not a big issue!

PR #506, opened 2019-01-29
[DEV APPROVED] Ensure filtering query returns unique pages

margOnline:
[TP 10209](https://moneyadviceservice.tpondemand.com/restui/board.aspx?#page=board/5682182231901749200&appConfig=eyJhY2lkIjoiRkRBMzk3RTBGNjU3NDg1RDM4QzczMTJFNzI5MDFBN0MifQ==&searchPopup=userstory/10209)

#### Bug
When searching documents on the evidence hub by multiple values for Topics or Client Groups, duplicate results are being displayed.

#### Solution
We execute a very complicated sql query in order to allow the user to filter/search documents by keyword, document type and/or various filters. Amend the query to ensure it only returns distinct results.

#### Technical detail
`DocumentProvider` builds the query, which only gets executed in the controller. We need to force the query to load to ensure the mix of raw sql and rails's `activerecord` methods used in the query, are performed in the correct order. 

Specifically, forcing the load ensures the `size` method acts on the results after the `distinct` is executed.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -374,6 +374,7 @@ Style/BlockDelimiters:
 Style/ConditionalAssignment:
   Exclude:
     - 'app/models/comfy/cms/category.rb'
+    - 'app/filters/document_provider.rb'
 
 # Offense count: 3
 # Cop supports --auto-correct.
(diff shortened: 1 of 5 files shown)


scruti:
Not going to officially review it as been pairing on it with @margOnline but want to highlight how nice is to add a failing test in order to get a starting point from where to fix things. :clap:

PR #509, opened 2019-02-18
[DEV APPROVED] [10142] - Ruby upgrade to 2.5.3

MarcoBgn:
[Tp Card](https://moneyadviceservice.tpondemand.com/entity/10142-upgrade-cms-repo-to-ruby-253)

## Context

- The main issue with this upgrade was the Cucumber suite. I had non-deterministic failures (~ 25% of the times: Deadlock in Mysql) which were quite tricky to track. After some digging it seems to be due to a mix of configuration and dependencies:

- `Capybara` was locked to a fairly outdated version, and this seems to be part of the issue. Main source of information [here](https://github.com/teamcapybara/capybara/issues/1089)

- `database_cleaner` was not configure to clean Around each feature which might have created concurrencies problems in the db (as far as I can see, a new spec trying to write while DC was still cleaning)

## Test results

- After these changes, I had a successful 15+ full runs on cucumber so I am quite confident to proceed (I am still testing and still have to encounter a failure, so even in case of a false positives chances are very low)

## Other small changes

- More info in the commits messages (i.e: I have added a fail-fast option to cucumber, so the dev doesn't have to wait for the full suite in case of issues)

### DIFF
--- a/features/support/env.rb
+++ b/features/support/env.rb
@@ -7,7 +7,11 @@ require 'capybara/poltergeist'
 require 'aruba/cucumber'
 require 'database_cleaner/cucumber'
 
-DatabaseCleaner.strategy = :truncation
+begin
+  DatabaseCleaner.strategy = :truncation
+rescue NameError
+  raise "You need to add database_cleaner to your Gemfile (in the :test group) if you wish to use it."
+end
 
 Around do |scenario, block|
   DatabaseCleaner.cleaning(&block)
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -62,7 +62,7 @@ GEM
       minitest (~> 5.1)
       thread_safe (~> 0.3, >= 0.3.4)
       tzinfo (~> 1.1)
-    addressable (2.5.2)
+    addressable (2.6.0)
       public_suffix (>= 2.0.2, < 4.0)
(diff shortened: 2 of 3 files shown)


features/support/env.rb
```diff
@@ -5,6 +5,17 @@
 require 'capybara/rails'
 require 'capybara/poltergeist'
 require 'aruba/cucumber'
+require 'database_cleaner/cucumber'
+
+begin
+  DatabaseCleaner.strategy = :truncation
+rescue NameError
+  raise "You need to add database_cleaner to your Gemfile (in the :test group) if you wish to use it."
```
tomas-stefano: Is this line needed if we include the database cleaner on the test group?

MarcoBgn: Not really, I believe it was included as a 'best practice' but happy to remove it if you feel it's redundant

MarcoBgn: Thanks for the review, removed 👍

MarcoBgn:
> Nice work, not sure all those commits are needed

Thanks @margOnline, will see if I can rebase

osharpe:
This PR has been super-seeded by https://github.com/moneyadviceservice/cms/pull/515. Closing this now

PR #515, opened 2019-02-19
Ruby upgrade to 2.5.3 and Jenkins

osharpe:
[TP](https://moneyadviceservice.tpondemand.com/entity/10142-upgrade-cms-repo-to-ruby-253)

## Context
* The main issue with this upgrade was the Cucumber suite. I had non-deterministic failures (~ 25% of the times: Deadlock in Mysql) which were quite tricky to track. After some digging it seems to be due to a mix of configuration and dependencies:
* `Capybara` was locked to a fairly outdated version, and this seems to be part of the issue. Main source of information [here](https://github.com/teamcapybara/capybara/issues/1089)
* `database_cleaner` was not configure to clean Around each feature which might have created concurrencies problems in the db (as far as I can see, a new spec trying to write while DC was still cleaning)

## Test results
* After these changes, I had a successful 15+ full runs on cucumber so I am quite confident to proceed (I am still testing and still have to encounter a failure, so even in case of a false positives chances are very low)

## Other changes
* More info in the commits messages (i.e: I have added a fail-fast option to cucumber, so the dev doesn't have to wait for the full suite in case of issues)
* Use Jenkins Base image
* Add Danger Gem
* Add Brakeman Gem


scruti:
Edited description to add context about `#field_text` introduction to whoever comes to this PR with hope to understand why this method was added.

PR #519, opened 2019-05-28
Update README.md

msleiman:


davidtrussler:
Actually, I believe this URL is correct - though it assumes you are running the CMS locally to be valid.

